PR #7, opened 2022-10-26
ESM support

pavi2410:


pavi2410:
@yusukebe Can you please take a look here as well?

yusukebe:
Hi @pavi2410 !

Yes, I will do it after releasing the new version of Hono core package.

But, we may keep this `node-server` only supporting CommonJS ( do not make both ESM and CommonJS support ). Perhaps, I think it will not throw the error with the new version of Hono.

pavi2410:
@yusukebe I just migrated my project to Deno and I feel good about this 😂 Feel free to close this. Thanks for your awesome project!

yusukebe:
@pavi2410 

Yes! If you have a problem, feel free to create the issue!

marvin-j97:
@yusukebe Is there any rationale behind not supporting ESM? `hono` has ESM support, and Node 18 has too, why shouldn't the node server have ESM support? Looking at the source code, it's a 5 minute change.

yusukebe:
@marvin-j97 

If you can create ESM modules, but, I think supporting both CommonJS and ESM might be a bit troublesome. We would need to write some conditions into `package.json` and possibly add the extension when importing modules. If there aren't any significant issues, I'd prefer not to support ESM.

marvin-j97:
> @marvin-j97 
> 
> If you can create ESM modules, but, I think supporting both CommonJS and ESM might be a bit troublesome. We would need to write some conditions into `package.json` and possibly add the extension when importing modules. If there aren't any significant issues, I'd prefer not to support ESM.

It just makes little sense to me that Hobo is CommonJS and ESM, but the Node adapter isn't. It just prevents any Node user from going full ESM, which is very much possible these days. I would rather see it going full ESM than staying CommonJS.

yusukebe:
If we are to support ESM with this Node Adapter, I believe we should ensure compatibility with both CommonJS and ESM to maintain backward compatibility. This means that if we can support both, we will be able to support ESM. I'll try to tackle this later.

PR #9, opened 2022-11-28
fix if statement bug in server

tavvy:
This change fixes a bug with the handling of `HEAD` requests.

Previously we would never match on a `HEAD` request and thus incorrectly create a body to the request, which would throw an uncaught error

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -46,7 +46,7 @@ const getRequestListener = (fetchCallback: FetchCallback) => {
       headers: headerRecord,
     } as RequestInit
 
-    if (!(method === ('GET' || 'HEAD'))) {
+    if (!(method === 'GET' || method === 'HEAD')) {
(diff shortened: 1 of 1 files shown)


tavvy:
spotted another small bug.
You can test with a `curl -I <local-endpoint>` and get a thrown error.

Perhaps it would be a good idea in a follow up change to catch and handle errors thrown by the fetchCallback

```
        let res;
        try {
          res = (await fetchCallback(new Request(url.toString(), init)));
        } catch {
          res = new Response(null, { status: 50x })
        }
```

yusukebe:
Hi @tavvy !

This is a very elementary bug! Thank you for your fix. I'll merge it.

> Perhaps it would be a good idea in a follow up PR to catch and handle errors thrown by the fetchCallback

Ah, you are right. It's a good idea.

PR #11, opened 2022-12-01
fix bug with encoded responses

tavvy:
This change fixes how we write out a response body so that we can support encoded responses

Previously we would cause a content encoding error by calling `res.text()` on responses even if they are encoded (`br`,`gzip` etc). Now we check if the response has the `content-encoding` header and use the read/write stream instead if it does.


yusukebe:
Hi @tavvy !

Thanks, it's a good fix. You are right, we have not cared about encoded response. OK, I'll merge it!

PR #19, opened 2023-01-24
global.crypto fix for Node 19.0

doge1338:
With Node.js 19.0, the Web Crypto API is now accessible directly from the global object. global.crypto is not writable so trying to assign to it crashes the program. This is a simple fix for it.


yusukebe:
Hi @doge1338 !

Thanks. This is a good fix! I'll merge it now.

PR #23, opened 2023-02-23
feat: add adapter for use with nextjs node runtime

hagishi:
This PR implements an adapter to enable the use of Hono on the Next.js Node runtime.


yusukebe:
Hi @hagishi !

Sounds good! The following points concern me.

*  It would be better to make it a Next.js adapter like Hono's core.
* Actually it should be renamed `src/next.ts` to `src/adapter/nextjs.ts`, but that would make `package.json` more complicated, so `src/nextjs.ts` is fine.
* We should not `export` it from `src/index.ts` for those who do not use Next.js.
* I think`import { handle } from @hono/node-server/nextjs` is better.

Consider these.

hagishi:
@yusukebe 

Thank you for response.
I fixed the code based on your feedback!! 🥇

yusukebe:
Thanks! I'll check it later.

yusukebe:
Hi @hagishi !

I've tried your branch with Next.js. That's great!

One thing. We want to pass the "path" to `handle` as the second argument.

```ts
app.get('/hello', (c) => {
  return c.json({
    message: 'Hello',
  })
})

export default handle(app, '/api') // the handler will be `/api/hello`
```

The implementation of the core is below:

https://github.com/honojs/hono/blob/bb929b44cfb611fea19c2c80394d4beef45078e8/src/adapter/nextjs/handler.ts#L9-L12

Could you see it?

hagishi:
Hi @yusukebe 

We've modified to accept path arguments in the handle.

I have one question.
@hono/node-server package dependent on hono v2. Would it be better to implement it using v3？

yusukebe:
@hagishi 

Thanks! Great work!

> @hono/node-server package dependent on hono v2. Would it be better to implement it using v3？

Yes, we should bump up the Hono to v3 (but this should be as "devDependencies").

karlhorky:
Should a test also be added for Next.js 13+ [Route Handlers](https://beta.nextjs.org/docs/routing/route-handlers) which use the new App Router? (in the `app/` directory)

PR #24, opened 2023-03-04
Must use relative paths

mindplay-dk:
Added documentation for `path`, which appears to require relative paths.

It was not obvious for someone porting a project with absolute paths from Express - the error messages were confusing:

```
Static file: ./home/mindplay/workspace/try-hono/public/index.css is not found
```

At a glance, it looks like the file exists - but the `.` at the beginning of course makes it relative to the current working directory.

It's not clear to me whether this is intentional, for platform portability perhaps? Most other Node servers would accept an absolute path - and this middleware only works on Node. But you might want relative paths in Hono projects as a convention?

If this is **not** intentional, you should reject this PR!

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,8 @@ import { serveStatic } from '@hono/node-server/serve-static'
 app.use('/static/*', serveStatic({ root: './' }))
 ```
 
+Note that `root` must be *relative* to the current working directory - absolute paths are not supported.
+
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @mindplay-dk !

Thank you for the PR. Indeed, it might confuse users.

I'll merge it!

PR #25, opened 2023-03-07
feat: use readable stream for the request body

tangye1234:
Using a ReadableStream instead of buffer.concat provides an improvement in several ways:

1. According to web standards, the request.body should either be null or an instanceof ReadableStream, not Buffer (Uint8Array).
2. In a Hono node server, it's important not to consume the incoming message body until it's actually used.
3. Using a ReadableStream is especially beneficial when clients post large bodies, as it allows the Hono app to be built quickly by the node server.

### DIFF
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -23,12 +23,15 @@ export const getRequestListener = (fetchCallback: FetchCallback) => {
     } as RequestInit
 
     if (!(method === 'GET' || method === 'HEAD')) {
-      const buffers = []
-      for await (const chunk of incoming) {
-        buffers.push(chunk)
-      }
-      const buffer = Buffer.concat(buffers)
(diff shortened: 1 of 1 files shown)


tangye1234:
Using this alternative approach might suffice as well. I discovered that @remix-run/web-fetch's Request object can convert a Node.js stream into a ReadableStream through the Stream.on('data', ...) method, which is intended to cater to older versions of Node.js.

However, in terms of the underlying mechanics, it would be more advantageous to use a ReadableStream, given that our Node server is an adapter based on Node 16. Employing async iterators appears acceptable and comes across as more lucid to developers.

yusukebe:
Hi @tangye1234 !

Great work! I also think it's better to use `ReadableStream`. Let's go with it!

PR #26, opened 2023-03-09
fix middleware link

e-jigsaw:
I wasn't sure which URL was correct, but it 404


yusukebe:
#62 has closed this problem.

PR #28, opened 2023-03-31
feat: add stream support and fix some bugs

tangye1234:
## Currently, c.body(stream) is not very friendly to stream support.
- Stream transfer;
- Potential crashes;
- Unable to cancel stream flow upon user abort request.

## Modify tsconfig es target to 2022 to support Error cause
- Only Error types are accepted by outgoing.destroy, but the thrown exception may not be an Error. Use es2022 cause to solve.

## hono throws a TimeoutError exception and returns a 504 instead of a 500 error.

### DIFF
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -38,12 +38,22 @@ export const getRequestListener = (fetchCallback: FetchCallback) => {
 
     try {
       res = (await fetchCallback(new Request(url.toString(), init))) as Response
-    } catch {
+    } catch (e: unknown) {
       res = new Response(null, { status: 500 })
+      if (e instanceof Error) {
+        // timeout error emits 504 timeout
+        if (e.name === 'TimeoutError' || e.constructor.name === 'TimeoutError') {
+          res = new Response(null, { status: 504 })
+        }
+      }
     }
 
     const contentType = res.headers.get('content-type') || ''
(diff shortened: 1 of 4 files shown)


src/listener.ts
```diff
@@ -55,12 +65,34 @@ export const getRequestListener = (fetchCallback: FetchCallback) => {
...
+         * Else if transfer-encoding is set, we assume that the response should be streamed.
+         * Else if content-length is set, we assume that the response content has been taken care of.
+         * Else if x-accel-buffering is set to no, we assume that the response should be streamed.
+         * Else if content-type is not application/json or text/* but can be text/event-stream,
+         * we assume that the response should be streamed.
+         */
+        if (
+          contentEncoding ||
+          transferEncoding ||
+          contentLenegth ||
```
yusukebe, line 80: Typo? It should be `contentLength`

yusukebe:
Hi @tangye1234 !

Looks good! One thing. I've noticed there is a typo.

tangye1234:
> Hi @tangye1234 !
> 
> Looks good! One thing. I've noticed there is a typo.

Sure

yusukebe (approved):

yusukebe:
Thanks!

PR #34, opened 2023-04-18
fix: add TransformStream to global and re-implement stream pipe

tangye1234:
- Rewrite the implementation of `Stream Pipe`, taking inspiration from nodejs' `stream.Readable.fromWeb`, which offers better support for backpressure and highWatermark option.

- Additionally, TransformStream has been added to globals.

### DIFF
--- a/src/globals.ts
+++ b/src/globals.ts
@@ -2,6 +2,9 @@ import crypto from 'node:crypto'
 import {
   ReadableStream as NodeReadableStream,
   WritableStream as NodeWritableStream,
+  TransformStream as NodeTransformStream,
+  TextDecoderStream as NodeTextDecoderStream,
+  TextEncoderStream as NodeTextEncoderStream,
 } from '@remix-run/web-stream'
 
 import { atob, btoa } from './base64'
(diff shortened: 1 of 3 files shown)


tangye1234:
This patch aims to address a potential memory leak that may occur when pushing large streams into the server response.

The server response features a backpressure mechanism, which causes the upstream to read data more slowly, thereby preventing stress on the response queue.

However without this patch, it can result in a significant increase in memory usage, as the sink may not have the capacity to handle the increased load efficiently.

tangye1234:
https://stackblitz.com/edit/node-mvpjwe?file=index.js
This is a demo to demonstrate that memory leak during huge stream processing.

You event cannot close the server, cause too many objects stuck in the response queue which prevent application to stop.
If you just arange the program a few memory, it will be finally OOM.

yusukebe:
OK, I'll check it later.

yusukebe:
Hi @tangye1234 !

Great work!

I tested your demo on my local machine, and indeed, the current version has memory leaks. However, with this PR, it doesn't consume a large amount of memory. Also, the implementation looks good.

I'll go ahead and merge it.

PR #38, opened 2023-04-21
feat: add `createServer` option

### LINKED ISSUE #37: SSL support
Hey! 
I'm trying to add SSL to our dev env but can't seem to access this line in an elegant way. 
I assume that by swapping that out with the `createServer` from `node:https` would do the trick.
https://github.com/honojs/node-server/blob/68b2475e45da38e9acdd79538c7c3cd235d6488a/src/server.ts#L12

A solution could be either default to `node:https` or expose `getRequestListener` so that we can create our own `serve` with https.

What do you think?

yusukebe:
This PR introduces the `createServer` option for both `serve` and `createAdaptorServer`. With this option, you can use the `createServer` function imported from `node:https`. This feature enables SSL support.

```ts
import { createServer } from 'node:https'
import fs from 'node:fs'

//...

serve({
  fetch: app.fetch,
  createServer: createServer,
  serverOptions: {
    key: fs.readFileSync('test/fixtures/keys/agent1-key.pem'),
    cert: fs.readFileSync('test/fixtures/keys/agent1-cert.pem'),
  },
})
```

Resolve #37


yusukebe:
Hi @tangye1234 @schonert !

I believe this is a good feature. You don't need to go into detail, but could you please review it?

src/server.ts
```diff
@@ -9,6 +9,7 @@ installGlobals()
 export const createAdaptorServer = (options: Options): Server => {
   const fetchCallback = options.fetch
   const requestListener = getRequestListener(fetchCallback)
+  const createServer = options.creteServer || createServerHTTP
```
Skye-31, line 12:
```suggestion
  const createServer = options.createServer || createServerHTTP
```

src/types.ts
```diff
@@ -12,6 +13,7 @@ export type Options = {
   port?: number
   hostname?: string
   serverOptions?: Object
+  creteServer?: typeof createServer
```
Skye-31, line 16: Likewise 🙂

README.md
```diff
@@ -52,13 +52,33 @@ Open `http://localhost:3000` with your browser.
...
+```ts
+import { createServer } from 'node:https'
+import fs from 'node:fs'
+
+//...
+
+serve({
+  fetch: app.fetch,
+  createServer: createServer,
+  serverOptions: {
```
schonert, line 75: I'd suggest leaving the `serverOptions` out by simply using the createServer as a function. Not seen the `serverOption` documented or a similar pattern used by other adaptors. Could potentially be a breaking change tho.

```ts
serve({
  fetch: app.fetch,
  createServer: (options = {}) => {
    return createServer({ 
        ...options, 
        key: fs.readFileSync('test/fixtures/keys/agent1-key.pem'),
        cert: fs.readFileSync('test/fixtures/keys/agent1-cert.pem'),
      })
    },
})
```

schonert:
Brilliant! I only had one comment :)

tangye1234 (approved):
This functionality can indeed resolve the HTTPS server problem, and the API suits me well. Looking ahead, it may require additional instrumentation like OpenTelemetry, so exposing the server creation option is an effective means of facilitating integration.

tangye1234 (approved):

README.md
yusukebe: Okay! I think you are right.

yusukebe: Ah, if we don't use `serverOptions`, we have to pass `requestListener` to createServer as an argument:

```ts
  const server = createAdaptorServer({
    fetch: app.fetch,
    createServer: (options = {}, requestListener: RequestListener) => {
      return createServerHTTPS(
        {
          ...options,
          key: fs.readFileSync('test/fixtures/keys/agent1-key.pem'),
          cert: fs.readFileSync('test/fixtures/keys/agent1-cert.pem'),
        },
        requestListener // <---
      )
    },
  })
```

It's a little bit annoying, so, keep using `serverOptions`.

PR #40, opened 2023-05-01
Feat/node18

tangye1234:
PR information:

1. replace @remix-run/* / atob / btoa / Blob / FormData polyfills with node native ones
2. retain only the fetch rewrite with compress opted-out + crypto load which is nothing to do with polyfill
3. use half duplex in hono.fetch init args when body is `ReadableStream`
4. use native `Readable.fromWeb` / `Readable.toWeb` for stream convertion
5. use node native Headers.prototype.getSetCookie for response headers with set-cookie
6. add `engines` to package, to ensure node version at least 18
7. add compress test case


yusukebe:
Hi @tangye1234!

From what I can see, this code looks perfect. It doesn't use any `@remix-run/*` polyfills and relies only on native Node.js APIs.

I've measured the performance of this Node.js adapter and compared it to Express.

Speed for a "Hello World" application:

SS

Bundle size for a "Hello World" application:

SS

Hono with the Node.js adapter is slightly faster than Express and has a much smaller bundle size. These results align with our expectations.

I'd like to proceed with this implementation.

One thing we need to consider is versioning. Merging this PR means we will no longer support Node.js versions below 16. If the current implementation is sufficient as a Node.js adapter, we can release it as v1.0.0, allowing for breaking changes with a major version bump.

Anyway, if there are no issues, please direct this PR to the "[next](https://github.com/honojs/node-server/tree/next)" branch. I'll merge it once it's ready. Thank you!

PR #41, opened 2023-05-04
Feat/node18

tangye1234:
PR information:

* replace @remix-run/* / atob / btoa / Blob / FormData polyfills with node native ones
* retain only the fetch rewrite with compress opted-out + crypto load which is nothing to do with polyfill
* use half duplex in hono.fetch init args when body is ReadableStream
* use native Readable.fromWeb / Readable.toWeb for stream convertion
* use node native Headers.prototype.getSetCookie for response headers with set-cookie
* add engines to package, to ensure node version at least 18
* add compress test case


yusukebe:
Hi @tangye1234 !

Thank you! I'll merge it.

PR #42, opened 2023-05-04
decode the path for serve-static

ff2400t:


yusukebe:
Thank you! I'll merge it.

PR #53, opened 2023-05-22
Fix/node18 parse set cookie

tangye1234:
This PR is to address the issue of parsing cookies in earlier Node.js 18 environments, a polyfill is implemented.

The most recent Node.js version 18 has already adopted the getSetCookie web standard. However, several versions earlier lack this capability internally. I've utilized a parsing method I've used before to parse the comma-separated cookie string, which TBH is significantly faster than a popular package called `set-cookie-parser`(about 7x faster with benchmarks being PRed together).

The test cases are more extensive than initially anticipated, with most parsing cases originating from `set-cookie-parser` which I must thank for. These cases have effectively covered the majority of potential issues.


yusukebe:
Hi @tangye1234 !

Thank you very much to creating the PR. I'll comment on later!

yusukebe:
Hi @tangye1234!

Great job! We still have a few things to do.

### 1. CI / GitHub Actions

This line should be `18.0.0` instead of `18.x` for testing:

https://github.com/honojs/node-server/blob/9783cc774f7550343ebbde8a3e27ab0d0d7549da/.github/workflows/ci.yml#L14

### 2. Avoid dynamic `require()`?

In my environment or in CI with the above fix, the tests will fail because it cannot resolve the path for this `require()`:

https://github.com/tangye1234/honojs-node-server/blob/a84f53eb385613ee03b1bd31712328ce635e2444/src/listener.ts#L59-L66

I understand we want to require the function dynamically to reduce the bundle size if it runs on Node.js, which supports `getSetCookie()`. However, I believe it's acceptable to use `parseSetCookie()` from `parse-set-cookie.ts`. Although the bundle size may slightly increase, it won't be overly large. What are your thoughts on this?

tangye1234:
According to my research, the earlest node.js version is `18.13.0`

There are three compatible issues about node 18.x
1. `FormData.prototype.forEach` is implemented in 18.4.0, earlier versions should be polyfilled.
2. `Headers.prototype.forEach` is implemented in 18.14.1, earlier versions should be polyfilled.
3. Pipeline stream from `response.body` to `outgoing` will throw error with EOF state abnormally, the isssue is fixed in 18.13.0, earlier version requires further investigation.

yusukebe:
Hi @tangye1234,

Thank you very much for your investigation and fixes.

To summarize, the API support is as follows:

* 18.x => 18.14.1+
* 19.x => 19.7.0+
* 20.x => 20.0.0+

Hmm,

polyfilling all functions, including `Headers.prototype.getSetCookie`, for earlier versions may be challenging. We might not even need to do it. We could simply document the supported Node.js versions. This might not be a significant issue since the latest versions of 18, 19, and 20 support it.

I know this has taken a considerable amount of your time, but how about closing the PR and clearly stating the supported versions in the README and on the web pages? Let me know your opinion.

tangye1234:
> Hi @tangye1234,
> 
> Thank you very much for your investigation and fixes.
> 
> To summarize, the API support is as follows:
> 
> * 18.x => 18.14.1+
> * 19.x => 19.7.0+
> * 20.x => 20.0.0+
> 
> Hmm,
> 
> polyfilling all functions, including `Headers.prototype.getSetCookie`, for earlier versions may be challenging. We might not even need to do it. We could simply document the supported Node.js versions. This might not be a significant issue since the latest versions of 18, 19, and 20 support it.
> 
> I know this has taken a considerable amount of your time, but how about closing the PR and clearly stating the supported versions in the README and on the web pages? Let me know your opinion.

Actually that's exactly what I was thinking too.
We should state that in the document.

PR #62, opened 2023-07-15
Fix broken built-in middleware link

collinstevens:
the previous link targeted a category/grouping page which no longer exists, so instead target the top-most document within the category it used to link to


yusukebe:
@collinstevens 

Thanks! I'll merge it!

PR #63, opened 2023-07-15
feat(serve-static): support range requests

yusukebe:
This PR will support [HTTP range requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests) in Serve Static Middleware. Based on #18


yusukebe:
@Hoodgail 

I've created this PR, based on your PR #18. I've added some tests and additional logic. I'll merge this, attributing you as a co-author.

@tangye1234 

Could you please review this?

src/serve-static.ts
```diff
@@ -14,36 +13,85 @@ export type ServeStaticOptions = {
   rewriteRequestPath?: (path: string) => string
 }
 
-export const serveStatic = (options: ServeStaticOptions = { root: '' }) => {
-  return async (c: Context, next: Next): Promise<Response | undefined> => {
+const createStreamBody = (stream: ReadStream) => {
```
tangye1234, line 16: Here we can use `Readable.toWeb(ReadStream)` instead, thus we can fully support backpressure.

yusukebe: Okay. Now let us you `Readable.toWeb()`. Thanks!

yusukebe:
@tangye1234 

Updated!

tangye1234 (approved):

yusukebe:
Thanks! Merge now.

PR #69, opened 2023-07-15
Refactor imports to type imports

marvin-j97:
`type` imports were used in some places,  but not all


yusukebe:
Thanks! I'll merge it.

PR #70, opened 2023-07-16
feat: support ESM

yusukebe:
With this PR, Node Server adapter will support both CommonJS and ESM.


marvin-j97:
Great news!

You could probably save yourself the build script and just use https://github.com/egoist/tsup which uses esbuild and tsc internally.

yusukebe:
@marvin-j97 

That looks great. I'll try it later!

PR #93, opened 2023-11-12
perf: reduce overhead

tsctx:
Hi @yusukebe,
I saw your [post](https://twitter.com/yusukebe/status/1723327398379475034)  on X (Twiiter) and tried to optimize it.
For performance improvement:

1. use `writeHead` instead of `setHeader`.
2. Quit changing `ReadableStream` to `Readable`. (Do not call `Readable.fromWeb`) 
3. Do not use `Readable.pipeline`.
4. use `Response#arrayBuffer` instead of  `Response#text` and `Buffer.byteLength`
5. There is no need to use `Headers#getSetCookie` since it is already split.

In other words, this change reduces the overhead of the handle portion of the body.
- *main*
`incomingMessage => Request => handling => Response => ReadableStream => Readable => outgoingMessage`
- *this PR*
`incomingMessage => Request => handling => Response => ReadableStream => outgoingMessage`


tsctx:
@yusukebe 
If you don't mind, could you measure the benchmark?

yusukebe:
Hi @tsctx!

Super cool! The five changes you made all seem good. The tests passed, and some real-world examples with this version, including streaming responses, work fine.

The benchmark I'm taking is super simple, just a "Hello World" application:

```js
import { Hono } from 'hono'

const app = new Hono()
app.get('/', (c) => c.text('Hi'))

serve(app)
```

And I access it with [Bombardier](https://github.com/codesenberg/bombardier) using this command:

```txt
bombardier -d 10s --fasthttp http://localhost:3000
```

The results:

Screenshot 2023-11-12 at 20 39 41

It really has become faster! In this case, it does not use streaming, so we may have to test another pattern, but currently, this simple pattern is a good reference.

I'd like to merge it into the "main". Do you have any more changes?

src/util.ts
```diff
@@ -0,0 +1,39 @@
+import type { Writable } from 'node:stream'
```
yusukebe, line 1: Could you rename this file to `utils.ts`?

tsctx: Ok.

tsctx:
You may merge them.

> Hi @tsctx!
> 
> Super cool! The five changes you made all seem good. The tests passed, and some real-world examples with this version, including streaming responses, work fine.
> 
> The benchmark I'm taking is super simple, just a "Hello World" application:
> 
> ```js
> import { serve } from '@hono/node-server'
> import { Hono } from 'hono'
> 
> const app = new Hono()
> app.get('/', (c) => c.text('Hi'))
> 
> serve(app)
> ```
> 
> And I access it with [Bombardier](https://github.com/codesenberg/bombardier) using this command:
> 
> ```
> bombardier -d 10s --fasthttp http://localhost:3000
> ```
> 
> The results:
> 
> Screenshot 2023-11-12 at 20 39 41
> It really has become faster! In this case, it does not use streaming, so we may have to test another pattern, but currently, this simple pattern is a good reference.
> 
> I'd like to merge it into the "main". Do you have any more changes?

You may merge them.
If you benchmark your storming, please let me know the results!

yusukebe:
Thanks! Okay, let's ship it.

If you have any ideas for improvement, we'd be glad if you created an issue or PR!

PR #94, opened 2023-11-12
perf: small improvements

Yovach:
I think these changes will bring small performance improvements because :
- the `split` function call will stop when we've retrieved the 2 elements we're using. It also prevents people from putting unnecessary elements in the `Range` header, which can affect performance.
- Reduce multiple assignments on `start` and `end` variables
- Call the `ReadableStream.getReader` function only if the `Writable` is not destroyed.


yusukebe:
Hi @Yovach 

Thanks. This is a good fix. LGTM!

PR #95, opened 2023-11-13
perf: Reduce object generation and optimize performance.

usualoma:
This is an optimization in a way that does not use the web standard Request and Response, so it may be out of line with the implementation policy in this repository, but this change made it 1.6 times faster.

Currently, only the simplest response where a string is returned is optimized, but if we work hard with additional PRs, we can also optimize the pattern where streaming is returned.

#### app

```ts
import { serve } from '../node-server/dist/index.js'
import { Hono } from './dist/index.js'

const app = new Hono()
app.get('/', (c) => c.json({ hello: 'world' }))

serve(app)
```

#### main branch

image

#### perf/proxy-object branch

image


usualoma:
with 5d8da3b

image

yusukebe:
Wow, ultra fast!

I'll review this later.

@tangye1234, what are your thoughts on this? Are there any problems?

src/listener.ts
```diff
@@ -8,36 +8,182 @@ import { writeFromReadableStream } from './utils'
...
-  ) => {
-    const method = incoming.method || 'GET'
-    const url = `http://${incoming.headers.host}${incoming.url}`
+const globalResponse = global.Response
+const responsePrototype: Record<string, any> = {
+  getResponseCache() {
+    this.__cache = undefined
+    return (this.responseCache ||= new globalResponse(this.body, this.init))
+  },
+  get body() {
```
rafaell-lycan, line 17: This would make a huge difference for devs migrating from Express to Hono in the next few weeks. 🚀 

Would it be possible to also support a built-in middleware such as body-parser (or a how-to guide) via config? cc @yusukebe 

Thanks!

src/listener.ts
```diff
@@ -8,36 +8,182 @@ import { writeFromReadableStream } from './utils'
...
+}
 
-    let res: Response
+const requestPrototype: Record<string, any> = {
+  getRequestCache() {
+    return (this.requestCache ||= newRequestFromIncoming(this.method, this.url, this.incoming))
+  },
+  get body() {
+    return this.getRequestCache().body
+  },
```
rafaell-lycan, line 119: This would make a huge difference for devs migrating from Express to Hono in the next few weeks. 🚀 

Would it be possible to also support a built-in middleware such as body-parser (or a how-to guide) via config? cc @yusukebe 

Thanks!

usualoma:
Sorry, I added this after the review started. I would also like to include the following commits.

* f0738ea

In hono-base.ts, there is an `instanceof Response` condition, so I made it behave more like a Response.

https://github.com/honojs/hono/blob/864fae668c7505e523cfe9d57e4e0778ddaa2fe5/src/hono-base.ts#L302

*  dcfa945

When a `Respose` object is returned from `fetch` instead of a `Promise`, it returns a response synchronously without using `await` or `then`.

This produced the following results in my environment

スクリーンショット 2023-11-15 5 36 04

src/listener.ts
usualoma: Hi @rafaell-lycan!
Thanks for your comment.

Do you think this will cause problems? If so, please let us know what specific issues you see.

> This would make a huge difference for devs migrating from Express to Hono in the next few weeks. 🚀

usualoma:
Incidentally, if we proceed with this implementation, I would like to merge the following two additional PRs after this
* https://github.com/usualoma/node-server/pull/1
* https://github.com/usualoma/node-server/pull/2

tangye1234:
The main idea here is to reuse the `bodyinit` information when building a `response` object to avoid re-parsing the `response body` when doing `node outgoing`.

After all, when node res.end is directly used with a string, or when the body uses non-stream objects such as `FormData`, `URLSearchParams`, it is faster to convert to a string rather than blob/stream.

In summary, for text data that is already prepared in memory, if it does not take up much memory, there is actually no need to go through `await response.text()`. It is more suitable to use `res.end` for direct output.

Currently, this idea can be utilized in some way, but the current PR changes are significant and it does not cover all kinds of situations that can be optimized. The code structure needs improvement.

I suggest adding more tests and isolating the intent of the code as much as possible to make the logic partition of the code clearer. It would be better to match some comments.

tangye1234:
After reviewing the source code of my NodeJS’s Undici, I’ve identified where the performance losses occur:

Take `new Response('a pure string', init?)` as an example.

The construction of the Response mainly involves the construction of Headers and Body [extractBody](https://github.com/nodejs/node/blob/416333fd503970227522602d5728d1ab31c81e3c/deps/undici/src/lib/fetch/body.js#L33C10-L33C21)）.

The handling of `BodyInit` in `extractBody` is as follows:

Ultimately, we want to obtain two results: Body{source,stream,length} and Type:

* The Body resides on top of `Response[kState].body`.
> https://github.com/nodejs/node/blob/main/deps/undici/src/lib/fetch/response.js#L485


* Type is used as an alternative plan when `Content-Type` is absent.
> https://github.com/nodejs/node/blob/416333fd503970227522602d5728d1ab31c81e3c/deps/undici/src/lib/fetch/response.js#L490C69-L490C69

The parsed Body’s source includes resources already in memory, such as:

| BodyInit                   | Source                  | Stream                  | Length                     | Type                                                       |
|--------------------|------------------|-------------------|-------------------|--------------------------------------|
| string                       | string                   ｜proxy stream       ｜BufferLength        | text/plain                                                 |
| URLSearchParams | string                   ｜proxy stream       ｜BufferLength        | application/x-www-form-urlencoded  |
| Buffer                      | Uint8Array          ｜proxy stream         | BufferLength        | octect/stream                                         |
| DataView                | Uint8Array            | proxy stream         | BufferLength        | null                                                           |
| ArrayBuffer             | Slice Uint8Array  | proxy stream         | BufferLength        | null                                                           |
| FormData              ｜FormData             | special stream       | null or known       | multipart/form-data                               |
| Blob                       ｜Blob                      | Blob.stream           | Blob.size               | Blob.type                                                |
| Stream                    | null                        | Stream it self         | null                        | null                                                          |
| [AsyncIterator]      | null                        | From Stream          | null                        | null                                                          |

Upon the completion of the Response construction, this is the content that exists within `Response[kState]`. If we need to output to the NodeJS outgoing, we can enhance speed by avoiding new memory allocation and avoiding re-constructing Stream and Pipeline.

Our current implementation is: For streaming Response types, we read the Stream directly and then write it into outgoing. 

* This current process doesn’t induce significant performance losses for BodyInit types such as FormData / Blob / Stream / [AsyncIterator], since the original stream has not been fully read into memory and we cannot know its content in advance.

* But for BodyInit types such as string / URLSearchParams / Buffer / DataView / ArrayBuffer, we can know the content in advance, and we can avoid re-constructing Stream and Pipeline by directly writing the content into outgoing.

That is the true meaning of this PR.

Let's look at the performance comparison: why is the performance of current implementation so poor?

* We use streaming when in some cases. 
> https://github.com/honojs/node-server/blob/5b0c13ba1f0a0ffbf3b21740bcb1d67c90d9e405/src/listener.ts#L84C57-L84C57

* If it’s not stream-based, we use await `response.arrayBuffer()` and output directly.
> https://github.com/honojs/node-server/blob/5b0c13ba1f0a0ffbf3b21740bcb1d67c90d9e405/src/listener.ts#L89
And let's seek into the arrayBuffer' implementation:
> https://github.com/nodejs/node/blob/416333fd503970227522602d5728d1ab31c81e3c/deps/undici/src/lib/fetch/body.js#L537C9-L537C22
> https://github.com/nodejs/node/blob/416333fd503970227522602d5728d1ab31c81e3c/deps/undici/src/lib/fetch/util.js#L836

The method `await arrayBuffer()` is using still using streaming, whether the body has a solid `Source` or not. This is the reason why the performance is also so poor.

So, the improvement we could make is to check for a `Source` when outputting. If a `Source` is present, we directly output the `Source`, avoiding output through a `Stream`. However, we should also consider the size of the `Source` whhich is already extracted by the nodejs in `extractBody` in stored in `Response[kState].body`. If it’s too large, we should still use streaming output to avoid excessive Watermark pressure.

------------------------

After a while, maybe in my spare time, I can come up with a more refined approach and endeavor to resolve this issue.

tangye1234:
Finally I get a tricky code here:

```javascript
const kState = Reflect.ownKeys(new Response()).find(
  k => typeof k !== "string" && k.toString() === "Symbol(state)"
)

function getResponseBody(response) {
  if (!kState) return

  const state = response[kState]
  if (!state || !state.body) return

  return state.body
}
```

Let's see some examples:
image

You can easily get the source / stream / length
If you wanna get the type, just use `response.type` instead

Then you known how to make` Response` to `node outgoing` right~:)

tangye1234:
After extensive testing, I found that the `new Request` / `new Response` are the main culprits killing performance.
Too many large objects are created by nodejs Request / Response during benchmark, which also verified why `koa` is faster than `expressjs`.

The story is, in the past we deliberated over whether to use @remix-run/fetch or nodejs native fetch. Nodejs undici seemed to be heavy, slower than @remix-run/fetch, and it creates more internal objects which kills the performance.

However the performance differences were marginal. We deemed that a more conventional implementation would be neat and compatible with more cases. Hence we decided to stick with the native fetch. Now, we see a more *BRALLIANT* implementation is taking place here.

Personally I am not sure whether we should workaround it with this PR, By this, I mean, should we make extensive efforts to avoid using the native Request / Response constructor, aiming at incredibly high speeds for simple use cases like a hello world benchmark? This could disrupt the simplicity of the project, as sometimes magic is just that - magic.

The original purpose behind this PR was to shed off the conversion between nodejs and web standards, aiming to employ the simplest methods possible. Maybe yes, it’s appropriate to leverage this concept as `hono` node-server is designed to bridge this gap.

@yusukebe, the decision lies with you, I think it has reached a stage where there's not much coding work left, it’s more about the direction you want to take.

yusukebe:
@usualoma @tangye1234 

Super thanks for your great work. I'll consider this deeply in this weekend.

usualoma:
Hi, @tangye1234!

Thanks for your thoughtful and detailed review. I fully agree with your comments. (I also learned a lot from the comments about the Node.js Response object submitted before that.)

While there is no doubt that this PR will achieve a very significant speedup on small benchmarks, it will compromise the simplicity of the implementation. Whether we should focus our efforts there or not is a matter of judgment.

For my part, even if this PR were to be merged, it is the effort of the node-server implementation up to this point that makes the result viable. If the bodyInit is not a string or ReadableStream, this PR’s Response will fall back to the Node.js Response object, which requires the existing implementation.

As an aside, this is similar to the relationship between Hono's RegExpRouter and TrieRouter. Where RegExpRouter, which is complex and huge in implementation but very fast, cannot handle some routing, TrieRouter, a straightforward implementation, can take it as a fallback destination.

Anyway, I have no objection to leaving it to the decision of @yusukebe.

Best regards!

usualoma:
P.S.

If someone asks me if this PR is a "cheat against benchmark scripts?" I would answer "No" to that question.

This is because most requests expected to be accelerated in a website production environment will likely be GET requests that return either a string or a ReadableStream.  Since this PR is an optimization for those (which is annoying considering GraphQL), it will also contribute to performance gains for the production environment (although the effect will be more minor as the response size increases).

RegExpRouter performs amazingly well in the benchmark scripts. Still, it is also faster in the production environment for "lots of handler registrations" and "lots of middleware registrations" than other routers. These characteristics are similar to this case.

In short, the content of this PR should produce results consistent with Hono's policy of "speedy in benchmark scripts, but also fast in production environments.

yusukebe:
# Let's make Hono one of the fastest frameworks for Node.js!

For Hono, being **fast** and **small** are most important things. So, let's work on making Hono the fastest framework for Node.js!

I've deeply considered the PR by @usualoma and comments by @tangye1234, and after running the app with the Node.js Adapter with this PR, I've decided to proceed with it.

The reason is simple: it makes Hono faster. My investigation revealed that the implementation isn't overly complex or "magical". The logic is simple: it makes operations that aren't immediately necessary lazy. This approach is already applied in some parts of Hono's core.

## How much faster will Hono be with this PR?

I'm comparing it with other frameworks using the scripts:

* Express
* Fastify
* Polka
* Hono with Node.js Adapter v1.2.3
* Hono with Node.js Adapter with this PR 
* Hono on Bun

The scripts you can use: https://gist.github.com/yusukebe/895f2ff55014c10f3e249a87e592ca49

One test is accessing `/`, returning plain text, and the other is returning a JSON response, as used in [Fastify's Benchmarks](https://github.com/fastify/benchmarks).

According to these tests, Hono is either the fastest or the second behind Fastify. We need to document these results, so I will do that.

## Next Actions

@usualoma

Could you run the benchmarks and look for more tuning points?

I'll review and comment later.

@tangye1234 

Your contribution so far is appreciated. I hope you will continue to do so. Thank you.

src/listener.ts
```diff
@@ -8,100 +8,268 @@ import { writeFromReadableStream } from './utils'
...
+        ? 504 // timeout error emits 504 timeout
+        : 500,
+  })
+
+const responseViaResponseObject = async (
+  res: Response | Promise<Response>,
+  outgoing: ServerResponse | Http2ServerResponse
+) => {
+  try {
+    res = await res
```
yusukebe, line 185: Can we make it does not `await` if it's not necessary?

src/listener.ts
```diff
@@ -8,100 +8,268 @@ import { writeFromReadableStream } from './utils'
...
-    for (let i = 0; i < len; i += 2) {
-      headerRecord.push([incoming.rawHeaders[i], incoming.rawHeaders[i + 1]])
-    }
+function newResponse(this: Response, body: BodyInit | null, init?: ResponseInit) {
+  Object.assign(this, {
+    status: init?.status || 200,
+    __body: body,
+    __init: init,
+    __cache: [body, (init?.headers || {}) as Record<string, string>],
+  })
```
yusukebe, line 76: Hono does not return a `Content-Type` header when returning a response with a metho like `c.text('hi')`.

We have to set `Content-Type` as `text/plain;charset=UTF-8` if it's not set, at here for somewhere.

yusukebe:
PS.

As we have discussed, being fast in benchmarks is not the same as being fast in real world. But being fast in benchmarks is good metrics, and as @usualoma said, in many cases this improvement will improve performance in the real world.

yusukebe:
The results on my machine:

Screenshot 2023-11-18 at 16 19 19

yusukebe:
The results for `/json`. Fastify is fastest.

Screenshot 2023-11-18 at 16 27 39

src/listener.ts
yusukebe: @usualoma Ah, you already did it!
https://github.com/usualoma/node-server/pull/2/files

usualoma:
Thank you. I will now merge the following and do some refactoring
https://github.com/usualoma/node-server/pull/1
https://github.com/usualoma/node-server/pull/2

I will ask for a re-review here when I am done.

yusukebe:
@usualoma

If you're interested, you might want to check out this repository: https://github.com/the-benchmarker/web-frameworks

This is another popular benchmark. Hono is tested on Bun, but the results show it as slow, because it might be running with a single thread for Bun. All frameworks for Bun don’t seem to perform well in this benchmark.

I think Hono has the potential to be the fastest among JavaScript frameworks.

usualoma:
> Could you run the benchmarks and look for more tuning points?

The/json performance is so good with fastify because the serializer is prepared in advance. This is indeed a good idea.

The npm that builds the serializer can be found here, which will also speed up Hono's /json. (It's already fast, though, and it isn't easy to stabilize the benchmark)
https://github.com/fastify/fast-json-stringify

I have seen improved performance with the following code
```ts
import fastJson from 'fast-json-stringify'
import { serve } from '../node-server/dist/index.mjs'
import { Hono } from './dist/index.js'

const stringify = fastJson({
    type: 'object',
    properties: {
        hello: {
            type: 'string'
        }
    }
})

const app = new Hono()
app.get('/', (c) => c.text('Hi'))
app.get('/json', (c) =>
    new Response(stringify({ hello: 'world' }), {
        headers: {
            'content-type': 'application/json; charset=UTF-8'
        }
    })
)

const port = 3001
console.log(port)

serve({
    fetch: app.fetch,
    port
})
```

Creating a custom serializer is not that difficult, and if it is a straightforward object, it can be implemented as follows, making it faster. However, supporting all kinds of schemas would be a big job.

```ts
import { serve } from '../node-server/dist/index.mjs'
import { Hono } from './dist/index.js'

const jsonEscapeRe = /["\\\n\r]/g
const isJsonEscapeRequired = (str) => jsonEscapeRe.test(str)

const buildSerializer = (schema) => {
    const list = []
    const keys = []
    for (const [k, v] of Object.entries(schema.properties)) {
        if (v.type === 'string') {
            list.push(`"${k}":"`)
            keys.push(k)
        }
    }
    list[0] = '{' + list[0]
    list.push('"}')
    return (data) => {
        let res = list[0]
        for (let i = 0; i < keys.length; i++) {
            const value = data[keys[i]]
            res += (isJsonEscapeRequired(value) ? JSON.stringify(value).replace(/^"|"$/g, '') : value) + list[i + 1]
        }
        return res
    }
}
const serializer = buildSerializer(
    {
        type: 'object',
        properties: {
            hello: {
                type: 'string'
            }
        }
    }
)

const app = new Hono()
app.get('/', (c) => c.text('Hi'))
app.get('/json', (c) =>
    new Response(serializer({ hello: 'world' }), {
        headers: {
            'content-type': 'application/json; charset=UTF-8'
        }
    })
)

const port = 3001
console.log(port)

serve({
    fetch: app.fetch,
    port
})
```

Considering the possibility of introducing this to Hono, it would be nice to be associated with jsonT or replaced somehow. However, it sounds like a challenging task.

yusukebe:
@usualoma

This topic is related to this issue: https://github.com/honojs/hono/issues/1706.

Indeed, the idea of changing the serializer for JSON is interesting. Building the schema during the route registration phase is a good for optimization. It might be different, but Elysia uses a similar approach for request JSON parsing.

Ideally, if we could provide a JSON serializer helper, allowing users to create their own serializers by defining a schema, and then use it in `c.json()` as a custom serializer, that would be great. The best goal would be to support `c.jsonT()`. However, as you mentioned, it might be quite challenging to implement.

usualoma:
This is off the topic of this PR, but the pattern of using a custom serializer with a specified schema is very smart. It can prevent incorrect information from being published.

```ts
const serializer = buildSerializer(
    {
        type: 'object',
        properties: {
            hello: {
                type: 'string'
            }
        }
    }
)
serializer({
  hello: "Hono",
  user: { id: 1, password: "Oooops" },
}) // => {"hello":"Hono"}
```

src/listener.ts
usualoma: Add default content-type by 2df0dd8d0 and 2bed5ba1f2e38

usualoma:
My work is done.

Would you please review this?
@yusukebe @tangye1234 

Although only some comments were added, 661a123 made the Request and Response definitions into independent files for better visibility.
The function names are descriptive throughout, so I don't think we will have trouble understanding them.

As for the tests, I have added some for the Request and Response fallback pattern. The existing tests will cover the others.

PodaruDragos:
@usualoma  @yusukebe you guys are doing a fantastic job !! Keep up the good work 🥇

src/listener.ts
yusukebe: Thanks!

test/server.test.ts
```diff
@@ -138,27 +148,81 @@ describe('Request body', () => {
...
-  it('Should return HTML', async () => {
-    const res = await request(server).get('/html')
-    expect(res.status).toBe(200)
-    expect(res.headers['content-type']).toMatch(/text\/html/)
-    expect(res.text).toBe('<h1>Hello!</h1>')
+  describe('Fallback to global.Response', () => {
+    const app = new Hono()
+
+    app.get('/json-blob', async () => {
+      return new Response(new Blob([JSON.stringify({ foo: 'blob' })]), {
```
yusukebe, line 200: Nice tests!

yusukebe (approved):
@usualoma 

LGTM! Awesome work!

yusukebe:
@tangye1234

Sorry to ping you again, but could you review this? If I don't hear back from you soon, I'll go ahead and merge it.

tangye1234:
I am testing an implementation, which can fully share the pros of the PR, but more cleaner, while also can satisfy these cases:

```javascript
// response name not changed
Response.name === 'Response'

// response prototype chain not changed
new Response() instanceof (await fetch('https://www.google.com')).constructor

// can only use new operator
expect(Response()).toThows()

// support Response static method
typeof Response.error === 'function'
typeof Response.json === 'function'
typeof Response.redirect === 'function'

// support other class to extends from Response
class NextResponse extends Response {}

// support ReadableStream / URLSearchParams / string / ArrayBuffer quick proxy(without constructing real Response)
```
But I don't know when it could be finished.

usualoma:
@tangye1234

Thank you for your important points! You are indeed correct.
In 04162de, I have ensured that the tests you have shown me succeed. How about this.
In my environment, adding this change did not degrade performance.

cc: @yusukebe

tangye1234:
I also did the same tests, completely depends on your branch.

https://github.com/tangye1234/honojs-node-server/tree/perf/proxy-object

I did not PR to here, cause I am respecting @usualoma's performant and beautiful job.

In my code snippets, there are 3 important things which should be clear:

1. Use symbol and private (#) members or methods as possible as I can, this will not leak the internal members and will never be conflicted to outside environments.
2. Polyfill `Response` in globals, not in response file itself, this will make all polyfills clearer and more controllable.
3. Use a wrapper `Request` class to construct the input rather than plain object. It is a similar approach to the `Response`, so the prototype chain can be easily maintained. The left things can be easily done, like parsing body or cookies.

Additionally, I did some work to support the stream / arraybuffer body and also support getting the native NodeJS Response state for a more directly proxy.

yusukebe:
Thanks, @tangye1234,

In this case, I prefer @usualoma's implementation (I think it's because I am familiar with reading @usualoma's code). Therefore, I'd like to proceed with this PR.

usualoma:
@tangye1234 
Thank you!

I have confirmed your branch.
https://github.com/tangye1234/honojs-node-server/tree/perf/proxy-object

> 1. Use symbol and private (#) members or methods as possible as I can, this will not leak the internal members and will never be conflicted to outside environments.

Yes, indeed, my branch was very messy in its implementation in this regard. It was fixed in e0c57d2.

It would also be reasonable to override global.Request via globals.ts. It was fixed in 5e28a91. However, this is a jest issue, since `global.Response = Response` does not override in the jest environment, so the current https://github.com/tangye1234/honojs-node-server/tree/perf /proxy-object, the test is running without Response being overwritten.

I think [getResponseInternalBody](https://github.com/honojs/node-server/blob/a8117ce9e4f6b1a2615dde1d1b4c2428b801cfa2/src/utils.ts#L13-L24) is an exciting implementation, it would help speed things up, reduce conditional branching in the current part of the following, and I think it is very attractive.
https://github.com/honojs/node-server/blob/main/src/listener.ts#L78-L81

However, since it depends on the internal implementation of Node.js, it is debatable whether this should be introduced to node-server, so it is better to make it another PR. Another PR is better because it will be clear that it is the result of @ tangye1234.

As for me, I would like to consider performance first here, so I think the content of this PR is the fastest at the moment, as indicated by https://github.com/honojs/node-server/pull/95#issuecomment-1820577597, so I would like to merge current HEAD. After that, improving with additional PRs as needed would be good.

tangye1234:
There is an important case you may missed.

```typescript
const app = new Hono()

app.get('/', c => {
  const isRealRequest = c.req.raw instanceof Request
  // should return true
  return c.json(isRealRequest)
})
```

In the listener, the incoming should be converted to a Request, which should be an instance of real nodejs Request.
This Request is not a polyfill, it is just a wrapper constructed by `IncomingMessage`.

usualoma:
@tangye1234

Thank you again.
I had not forgotten about that point, but I could not think of a situation where a confirmation like `c.req.raw instanceof Request` would be needed, so I thought that if there was such a use case, it could be handled in a later minor version. Nevertheless, as you point out, if we can handle it, it would be better to have it handled from the beginning.

It was fixed in a9c3cd2

yusukebe:
@usualoma 

Sorry for the delay! Is this ready to be merged? Okay??

I'll release the **v1.3.0** includes this PR after merging.

usualoma:
@yusukebe
Welcome back!
Yes, I believe we can merge the current branch's HEAD.

yusukebe:
Okay! Let's gooooooooooo!!

PR #96, opened 2023-11-15
fix: handle calling writable.end() correctly

ezg27:
Hi there! It looks like a bug was introduced in v1.2.2 with the addition of `writeFromReadableStream` in [this PR](https://github.com/honojs/node-server/pull/93). I was able to reproduce this on both Node v18 and v20.

With fast responses on a fast connection the problem isn't as noticeable, but if you're serving large responses (particularly if the response returned from a Hono handler is a ReadableStream that's slow to enqueue chunks), then `writable.end()` in `writeFromReadableStream` is never called. This leads to responses in the browser showing as 'Pending' indefinitely.

I've created a reproduction repo [here](https://github.com/ezg27/hono-node-server-repro) that demonstrates the issue - try adding a log inside the `if (done) {...}` statement in the `flow` function in `node_modules/@hono/node-server/dist/index.mjs` and you'll see it never gets logged. The reproduction returns an image response as a stream, with a short artificial delay inserted before each chunk is enqueued to better reproduce the problem.

The changes in this PR align more closely with [this example](https://nodejs.org/api/stream.html#event-drain) from the Node.js documentation, where a single-use `drain` listener is registered with `writable.once` each time `writable.write` returns `false`. This should ensure that write calls are properly paused + resumed and the stream read to the end with the writable closed successfully.

Let me know if you would prefer a different approach or if there's anything else you'd like me to add to this 🙂

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -8,14 +8,12 @@ export function writeFromReadableStream(stream: ReadableStream<Uint8Array>, writ
     return
   }
   const reader = stream.getReader()
-  writable.on('drain', onDrain)
   writable.on('close', cancel)
   writable.on('error', cancel)
   reader.read().then(flow, cancel)
   return reader.closed.finally(() => {
     writable.off('close', cancel)
     writable.off('error', cancel)
-    writable.off('drain', onDrain)
   })
   function cancel(error?: any) {
     reader.cancel(error).catch(() => {})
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @ezg27 !

Thanks for the PR and creating the project to reproduce the bug. Looks good. I'll merge it now!

sjc5:
This fixed an issue for me as well. Thank you!

PR #109, opened 2023-12-12
fix: Nested constructors should not cause an error even if ReadableStream is specified

usualoma:
Sorry, the fix in #108 was inaccurate.

Adding the following changes will eliminate the problem while maintaining high performance and more closely resembling the behavior of global.Response.

* If global.Response is not instantiated, #init itself is passed to the child.
* Instantiate global.Response immediately if it has already been instantiated


yusukebe:
@usualoma 

Thanks! Looks good to me. Let's ship it.

PR #115, opened 2023-12-28
fix: handle response that is not an instance of `Response`

### LINKED ISSUE #114: Handle an invalid response
It should handle an invalid response that is not an instance of `Response`:

```ts
app.get('/invalid', () => {
  return '<h1>Hello!</h1>'
})

serve({
  fetch: app.fetch,
  port: 3000
})
```

Currently, the application does not throw the correct error:

Screenshot 2023-12-27 at 22 03 57

yusukebe:
Fixes #114

Handling of responses returned from applications that are not `Response` instances, returning a proper response.

### DIFF
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -25,6 +25,10 @@ const handleResponseError = (e: unknown, outgoing: ServerResponse | Http2ServerR
     console.info('The user aborted a request.')
   } else {
     console.error(e)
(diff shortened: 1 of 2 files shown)


yusukebe:
@usualoma 

If you can, please review this!

usualoma:
@yusukebe LGTM!

PR #117, opened 2023-12-30
fix: don't use `instanceOf Response` when handling response

### LINKED ISSUE #116: Error: The response is not an instance of Response, but [object Object]
In https://github.com/honojs/node-server/pull/115, an explicit `instanceof Response` check was added. This breaks for libraries/integrations that use the [@whatwg-node/events](https://www.npmjs.com/package/@whatwg-node/events) ponyfill package. A prominent example is graphql-yoga which can no longer be used with hono & node-server because the response returned is an instance of [PonyfillResponse](https://github.com/ardatan/whatwg-node/blob/00a0dd315e4afd927870980ebbb574e6cda2f208/packages/node-fetch/src/Response.ts#L15) that does not pass the `instanceof Response` check.

Reverting to 1.3.3 resolves the problem.

yusukebe:
Fixes #116

In this PR, it avoids using `instanceOf Response` to determine if a response is right. This means that it will not throw an error when returning a ponyfill Response, e.g. `@whatwg-node/fetch`. The error message is now unfriendly, but this is the correct behavior.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
     "@types/jest": "^29.5.3",
     "@types/node": "^20.10.0",
     "@types/supertest": "^2.0.12",
+    "@whatwg-node/fetch": "^0.9.14",
     "eslint": "^8.55.0",
     "hono": "^3.11.7",
     "jest": "^29.6.1",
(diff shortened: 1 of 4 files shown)


yusukebe:
@usualoma Could you review this?

usualoma:
@yusukebe Thanks 👍

lorefnon:
@yusukebe  @usualoma Thank you.

PR #124, opened 2024-01-21
fix(request): handle "double dots" in URL

### LINKED ISSUE #123: Security vulnerability with serveStatic
Hello,
i recently discovered this library and I'm loving it so far. Thanks for the great work. ❤️ 

While working with the `serveStatic` method I noticed a behavior that I would describe as a security vulnerability.

Here's a simplified version of my project structure:
```
my-project/
├── node_modules
├── static/
│   └── global.css
├── src/
│   └── index.ts
├── package.json
└── .env
```

Here's a simplified version of my `src/index.ts` file. It's basically just copied from the [docs](https://hono.dev/getting-started/nodejs#serve-static-files).
```javascript
// src/index.ts
import { serveStatic } from '@hono/node-server/serve-static'

app.use('/static/*', serveStatic({ root: './' }))
```

After booting up the server I was able to get my CSS file by sending a GET request to `http://localhost:3000/static/global.css`, nice 👍.
To my surprise I was also able to get the content of my `.env` file by sending a GET request to `http://localhost:3000/static/../.env`. That seems like a security vulnerability to me. I would expect that the scope is restricted to the `static` folder. Am I missing something?

I'm happy to provide further information and create a pull request.

yusukebe:
Fixes #123

Since v1.3.0, we use our own Request object. This is great, but the `url` behavior is unexpected.

In the standard API, if the URL contains `..`, here called "double dots", the URL string returned by Request will be in the resolved path.

```ts
const req = new Request('http://localhost/static/../foo.txt') // Web-standards
console.log(req.url) // /http://localhost/foo.txt
```

However, our `url` as implemented below does not resolve double dots, so `http://localhost/static/.. /foo.txt` is returned.

https://github.com/honojs/node-server/blob/8cea466fd05e6d2e99c28011fc0e2c2d3f3397c9/src/request.ts#L43-L45

This causes vulnerabilities like #123 when using `serveStatic`.

So this must be fixed immediately. This PR solves that. The path will be checked on every request, which may have a performance impact but must be allowed for.


yusukebe:
Hi @usualoma !

I wanted to have this PR reviewed, but since it should be released soon, I am merging it. I think the behavior is fine, and if you can improve the performance, I would like you to create another PR.

Issue #127, opened 2024-01-23
Support compression gzip and deflate


ariskemper:
References https://github.com/honojs/hono/issues/2061

PR #128, opened 2024-01-24
feat [#127] : support compression gzip, deflate

ariskemper:
it fix issue #127 with using compress middleware on node.js https://github.com/honojs/hono/issues/2061

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,16 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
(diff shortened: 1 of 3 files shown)


.vscode/launch.json
```diff
@@ -0,0 +1,16 @@
+{
```
yusukebe, line 1: Is this file necessary?

yusukebe:
Hi @ariskemper !

Thanks so much. Looks great to me!

@usualoma Could you review this?

usualoma:
Hi @ariskemper 
Thanks for creating the PR!

But I think the problem with https://github.com/honojs/hono/issues/2061 is not with "node-server" but with "core". It should be fixed once the following PR is merged.
https://github.com/honojs/hono/pull/2080

cc: @yusukebe

Issue #122, opened 2024-01-19
Make Original `IncomingMessage` and `ServerResponse` Publicly Accessible


BeeeQueue:
Right now the original `IncomingMessage` from node [is being added to `Request#symbol(incomingKey)`](https://github.com/honojs/node-server/blob/main/src/request.ts#L82) but since it's using the symbol it is not possible to access afterwards.

I'm trying to see if it would be possible to use Hono in our applications at work and this would make it a lot easier as it allow us to write a wrapper around middleware for other frameworks or native node handlers.

i.e. `c.req.raw.incomingMessage`, `c.res.raw` or something similar

PR #90, opened 2023-11-08
Support upgrading requests from the adapter

Kyiro:
These changes allow the upgrading to something like WebSockets from within Hono itself. Feel free to request any changes or even deny this PR as I'm not sure if it fits. The only big issue is that I don't think it supports HTTP2 at all.
```ts
class UpgradeResponse extends Response {
    status = 101;
    
    constructor(body?: BodyInit | null, init?: ResponseInit) {
        super(body, init);
    }
}

const app = new Hono<{
    Bindings: {
        incoming: IncomingMessage;
        socket: Socket;
        head: Buffer;
    }
}>();
const wsServer = new WebSocketServer({ noServer: true });

app.get("/", (c) => {
    wsServer.handleUpgrade(c.env.incoming, c.env.socket, c.env.head, (ws) => {
        ws.send("Hello World!");
    });
    
    return new UpgradeResponse();
});

serve(app);
```
> This is only supposed to be an example of what a middleware should do as it's not too clean

Issue #80, opened 2023-09-02
Unable to access remoteAddr in route using node


shabbyrobe:
In my Deno applicaiton, I'm able to access the client's IP address on each request using some fairly convoluted, but nonetheless functional, means:

```js
const app = new Hono();
app.get("/", (ctx) => {
  return ctx.body(ctx.env.remoteAddr());
});

serve(async (request, connInfo) => {
  const remoteAddr = () => resolveRemoteAddr(connInfo.remoteAddr);
  const env = { remoteAddr };
  return app.fetch(request, env);
}, {
  hostname: "0.0.0.0",
  port: 3001,
});
```

This does not appear possible in an equivalent node application, using `honojs/node-server`. The function returned by `getRequestListener` has access to the `IncomingMessage` structure, which makes this available via `message.socket.remoteAddress`, but that is not itself made available to the `fetchCallback`, and the undici `Request` object that is passed to the route doesn't appear to provide access to it either.

Am I missing something?

yusukebe:
Hi @shabbyrobe,

At the moment, the Node.js Adapter doesn't provide support for accessing the remote address. This is due to the fact that the web standard API's `Request` object doesn't offer a method to access the IP address.

I believe it might be a good idea to introduce functions for retrieving the remote address via `env` variables. However, implementing this would require modifications to the Node.js Adapter itself.

shabbyrobe:
Hi @yusukebe, thanks for the context. Happy to try my hand at putting together a PR to add those functions to the adapter if you like? Where would you suggest might be the best place to add them?

yusukebe:
@shabbyrobe 

In my opinion, how about changing this line as follows: 

https://github.com/honojs/node-server/blob/48104d5c1a841e50d8aa090e7902cbe8a710fa90/src/server.ts#L8

pass `env` as the second argument to `fetchCallback`:

```ts
const env = {
  remoteAddress: incoming.socket.remoteAddress,
  //...
}
res = (await fetchCallback(new Request(url.toString(), init), env)) as Response
```

yusukebe:
@tangye1234 

If you have any opinion, please share it!

tangye1234:
> @tangye1234
> 
> If you have any opinion, please share it!

```typescript
const env = {
   // in nextjs, sometimes, incoming message is faked without socket
   remoteAddress: incoming.socket?.remoteAddress,
   // maybe useful
   httpVersion: incomming.httpVersion
}
```

[...]

PR #129, opened 2024-01-25
feat: Forward Raw `IncomingMessage` and `ServerResponse` Via `Env`

BeeeQueue:
Makes the raw `IncomingMessage` and `ServerResponse` instances accessible via `c.env` the [same way](https://github.com/honojs/hono/blob/main/src/adapter/aws-lambda/handler.ts#L158-L162) that [`hono/aws-lambda` does it](https://hono.dev/getting-started/aws-lambda#access-requestcontext).

After implementing it I saw that #90 does the same thing, but I thought I may as well submit it as it is smaller in scope.

```ts
type Bindings = {
  incoming: IncomingMessage
  outgoing: ServerResponse
}

const app = new Hono<{ Bindings: Bindings}>()

app.get('/', (c) => {
  console.log(c.env.incoming.url)
})
```

Related: #80 #90 #122

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,2 +1,3 @@
 export { serve, createAdaptorServer } from './server'
 export { getRequestListener } from './listener'
+export type { HttpBindings, Http2Bindings } from './types'
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -115,7 +115,7 @@ export const getRequestListener = (fetchCallback: FetchCallback) => {
     const req = newRequest(incoming)
 
     try {
-      res = fetchCallback(req) as Response | Promise<Response>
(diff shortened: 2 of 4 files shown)


yusukebe:
@BeeeQueue 

Looks great!

One thing, we could export the following `Bindings` types.

```ts
export type Bindings = {
  incoming: IncomingMessage
  outgoing: ServerResponse
}
```

I think users should import from `@hono/node-server`.

```ts
import type { Bindings } from '@hono/node-server'
```

cc: @Kyiro What do you think about this PR though I think you will say OK.

Kyiro:
looks good to me

yusukebe (approved):
Looks great!

yusukebe:
@BeeeQueue 

I'll merge this now. Thanks!

PR #130, opened 2024-01-25
feat: bump hono to v3.12.8, add customer err tests for compress middl…

ariskemper:


usualoma:
@ariskemper 
Thanks very much. 
It seems there was an error in the way I corrected the problem on https://github.com/honojs/hono/pull/2080.
I will correct it later.
Please let me put this PR on hold until then.

ariskemper:
@usualoma @yusukebe i refactored tests, fix UTF-8 issue and bump the version. Should be ready for merge.

usualoma:
@ariskemper 
Thank you very much!

Please remove the `console.log` statement.

Is there any reason why it should be `3.12.8` instead of `^3.12.8` on a dare? I think `^3.12.8` is better for me, because I think the cognitive load is high if only this part differs from the surroundings, although the difference is not so big since both are packages we manage ourselves.

ariskemper:
@yusukebe could you create a new release including this PR?

yusukebe:
@ariskemper 

Okay! But please run `yarn lint:fix && yarn format:fix`.

yarn.lock
```diff
@@ -2924,10 +2924,10 @@ hexoid@^1.0.0:
   resolved "https://registry.yarnpkg.com/hexoid/-/hexoid-1.0.0.tgz#ad10c6573fb907de23d9ec63a711267d9dc9bc18"
   integrity sha512-QFLV0taWQOZtvIRIAdBChesmogZrtuXvVWsFHZTk2SU+anspqZ2vMnoLg7IE1+Uk16N19APic1BuF8bC8c2m5g==
 
-hono@^3.11.7:
-  version "3.11.7"
-  resolved "https://registry.yarnpkg.com/hono/-/hono-3.11.7.tgz#e44f8aa3a18f19775304328bde6f559f7a02447f"
-  integrity sha512-TcfAq7IdipF+9coxnuzYlSSBXbm9mTyWjjagLCv/2ampboNcKJdi+XCK5G48mHQtpI5+9Rj3J4FfcGgw9vzIww==
+hono@^3.12.8:
```
yusukebe, line 2927: I don't think this `yarn.lock` change is necessary.

usualoma:
@ariskemper
Thanks again. Just sorry, but I would like to request the following.

* Changing the style of if statements is not considered a refactoring by us
    * This is a linter issue, so please don't change the if statement as a refactoring in PR
    * https://github.com/orgs/honojs/discussions/2108

ariskemper:
@usualoma i have just seen, that if, else is used in many places and many times could be written differently or simplified, but this change was more objective changing if, else to condition ? :

usualoma:
@ariskemper 
Thanks for the explanation! I am not familiar with eslint, but you are right, eslint may not do the conversion to ternary operator.

In any case, I don't think changing the style of if statements is something to be mixed into PR as a refactoring, since it is largely a matter of preference.

If you think such a rewrite in general would be useful, please create a separate PR for it and convince us by writing a rationale why you think it would be useful.

yusukebe:
Ternary is useful, but surely we should also specify in Linter whether ternary should be used or not.

https://eslint.org/docs/latest/rules/no-ternary

We can do this in ESLint with the `no-ternary` option. Later on I will change the ESLint rules to give an error when using this Ternary.

ariskemper:
@usualoma @yusukebe it actually shortens the code and remove identation, that's why i prefer it over single line block if, else statements.
Offcourse, we should have eslint rules to allow/disallow it, depends on agreed code guidelines. I prefer to do large PRs when refactor some codebase and it's better to improve code in smaller iterations.

usualoma:
### ternary

The ternary itself I don't dislike; I prefer to use it when returning.

```ts
return condition ? a : b
```

However, I think it is bad style to assign within ternary.

```ts
k === 'set-cookie' ? cookies.push(v) : (res[k] = v)
```



But sorry, about 0885eab, I don't think it's right to mix it in a PR that says "feat: bump hono to v3.12.8, add customer err tests for compress middl..."

ariskemper:
@usualoma could you explain more in detail, why do you think it should be used just when return a value and why do you think it's a bad style?

yusukebe:
Instead of disallowing all ternaries, it might be better to disallow a nested ternary. This is indeed very bad for readability.

Screenshot 2024-02-01 at 8 12 05

usualoma:
@ariskemper 
I will declare ahead of time that the following is my preference.

I think ternary should be used as an "expression (returning a value)". I don't think it should be used as a "statement" for side effects.

This is fine, since ternary is used as an "expression",
```ts
condition ? a : b
```

This is bad because it is a ternary used as a "statement".

```ts
k === 'set-cookie' ? cookies.push(v) : (res[k] = v)
```


But still, I don't think it is good that this discussion is taking place in a PR that says "feat: bump hono to v3.12.8, add customer err tests for compress middl..."

usualoma:
Once the rules are set, we will follow them, but for my taste, I find this example to be fine.
https://github.com/honojs/node-server/pull/130#issuecomment-1920151918

As for ternary, I think it is good when used as an "expression" and bad when used as a "statement" with the main purpose of side effects. Personally, however, I think that in many cases it is "fine to write it either way" and it is often better not to dare to subject it to refactoring.

usualoma:
This PR did an excellent job and led to fixing bugs in hono itself. Thanks to @ariskemper for creating it.
https://github.com/honojs/hono/pull/2080

I think 0885eab should be revert and merged as I think it does more than the job this PR should do.

I think it is better to discuss elsewhere whether 0885eab is needed or not.

yusukebe:
Thanks both.

Merging now.

PR #131, opened 2024-01-26
feat(listener): support custom error handler

marbemac:
Would ya'll be open to this kind of option? 

Open to alternative approaches, but basically what I'm after is a means for the consumer to be able to provide it's own error handler so that it can, for example, pass the error down the middleware chain in a Vite dev server. This allows vite to display the standard pretty error overlay on the client, etc.

See example of how it can be used in https://github.com/honojs/vite-plugins/pull/63 (this PR blocks that PR).

### DIFF
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -62,7 +62,9 @@ const responseViaResponseObject = async (
         res = await res
       } catch (err) {
         const errRes = await options.errorHandler(err)
-        if (!errRes) return
+        if (!errRes) {
+          return
+        }
         res = errRes
       }
     } else {
@@ -153,7 +155,9 @@ export const getRequestListener = (
       if (!res) {
         if (options.errorHandler) {
           res = await options.errorHandler(e)
-          if (!res) return
+          if (!res) {
+            return
+          }
         } else {
           res = handleFetchError(e)
         }
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -52,14 +52,14 @@ const responseViaCache = (
 const responseViaResponseObject = async (
   res: Response | Promise<Response>,
   outgoing: ServerResponse | Http2ServerResponse,
-  customErrorHandler?: (e: unknown) => void
+  options: { errorHandler?: (e: unknown) => void } = {}
 ) => {
   if (res instanceof Promise) {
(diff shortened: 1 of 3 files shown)


src/listener.ts
```diff
@@ -51,10 +51,19 @@ const responseViaCache = (
 
 const responseViaResponseObject = async (
   res: Response | Promise<Response>,
-  outgoing: ServerResponse | Http2ServerResponse
+  outgoing: ServerResponse | Http2ServerResponse,
+  customErrorHandler?: (e: unknown) => void
```
yusukebe, line 55: How about making the argument an object instead of a `customErrorHandler`? 

```ts
const responseViaResponseObject = async (
  res: Response | Promise<Response>,
  outgoing: ServerResponse | Http2ServerResponse,
  options?: {
    errorHandler?: (e: unknown) => void
  }
) => {
```

In that case, I would prefer a simple `errorHandler` in the `options`.

src/listener.ts
```diff
@@ -103,7 +112,10 @@ const responseViaResponseObject = async (
   }
 }
 
-export const getRequestListener = (fetchCallback: FetchCallback) => {
+export const getRequestListener = (
+  fetchCallback: FetchCallback,
+  customErrorHandler?: (e: unknown) => void
```
yusukebe, line 117: Same as above.

```ts
export const getRequestListener = (
  fetchCallback: FetchCallback,
  options?: {
    errorHandler?: (e: unknown) => void
  }
) => {
```

yusukebe:
Hi @marbemac !

This is an awesome feature! I have been wanting a feature like this. Please check my comments. Also, please add proper tests. Thanks.

marbemac:
Hi @yusukebe - alrighty, I've adjusted the function signature to accept the options object, and added some tests. LMK if the code style / test style works for you, or if you'd like any other changes!

yusukebe:
Hi @marbemac 

Please run:

```txt
yarn lint:fix && yarn format:fix
```

marbemac:
done

yusukebe:
Greeeat! I'll merge it and ship the new version. Thanks!

PR #132, opened 2024-01-29
refactor: more readable

jarqvi:
I've made some slight adjustments to enhance readability.

### DIFF
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -53,12 +53,11 @@ const responseViaResponseObject = async (
   res: Response | Promise<Response>,
   outgoing: ServerResponse | Http2ServerResponse
 ) => {
-  if (res instanceof Promise) {
-    res = await res.catch(handleFetchError)
-  }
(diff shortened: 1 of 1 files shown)


src/listener.ts
```diff
@@ -77,20 +76,30 @@ const responseViaResponseObject = async (
...
-        !regContentType.test(resHeaderRecord['content-type'] as string)
+        (accelBuffering && regBuffer.test(accelBuffering as string)) ||
+        !regContentType.test(contentType as string)
       ) {
         outgoing.writeHead(res.status, resHeaderRecord)
+
         await writeFromReadableStream(res.body, outgoing)
       } else {
         const buffer = await res.arrayBuffer()
         resHeaderRecord['content-length'] = buffer.byteLength
```
jarqvi, line 101: But I don't need to use the contentLength variable here
Because the content-length in resHeaderRecord should have changed.

yusukebe: This is needed.

yusukebe:
@jarqvi Thanks. Merge now.

PR #133, opened 2024-01-29
feat(request): Enable to create Request from light weight request.

usualoma:
This PR allows the following code to work, making it easy to replace the Request object in hono.

https://github.com/honojs/node-server/compare/main...usualoma:node-server:feat/create-request-from-request?expand=1#diff-e7c9480196e722de595dba06f56ad6601bce3d0e66651e7b77eea71f93be2ac3R55-R85

### DIFF
--- a/src/globals.ts
+++ b/src/globals.ts
@@ -1,9 +1,4 @@
 import crypto from 'node:crypto'
-import { Response } from './response'
-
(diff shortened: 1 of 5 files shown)


yusukebe (approved):
Looks good to me!

PR #135, opened 2024-01-31
feat(serve-static): mimes option for serve-static

ryuapp:
Related: https://github.com/honojs/hono/pull/2094

Waiting for Hono v4.


yusukebe:
@ryuapp 

This feature requires Hono v4. Since `@hono/node-server` has some v3 users, it may be better not to merge it now; if more users use v4, we can specify v4 or higher in `peerDependencies` and merge this PR.

Issue #107, opened 2023-11-30
Abort is not working.


buwon:
c.req.raw.signal does not work even when the client is disconnected.
It seems that there is no event handling for http.ClientRequest.

sjc5:
Noticed this as well.

PR #141, opened 2024-02-06
Catch when nodejs request is aborted

M4RC3L05:
When we receive a request, we start listening to the close event. When fired, we check if the request was aborted using `incoming.aborted`, if it was, we dispatch an abort event to the existing request abort signal.

Without this, when the client aborts the request, the signal on the request was not being called with the abort event.

Closes: https://github.com/honojs/node-server/issues/107

### DIFF
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -1,6 +1,6 @@
 import type { IncomingMessage, ServerResponse, OutgoingHttpHeaders } from 'node:http'
 import type { Http2ServerRequest, Http2ServerResponse } from 'node:http2'
-import { newRequest } from './request'
+import { getAbortController, newRequest } from './request'
 import { cacheKey } from './response'
 import type { CustomErrorHandler, FetchCallback, HttpBindings } from './types'
 import { writeFromReadableStream, buildOutgoingHttpHeaders } from './utils'
@@ -142,6 +142,13 @@ export const getRequestListener = (
     // `fetchCallback()` requests a Request object, but global.Request is expensive to generate,
     // so generate a pseudo Request object with only the minimum required information.
     const req = newRequest(incoming)
+
+    // Detect if request was aborted.
+    outgoing.on('close', () => {
+      if (incoming.destroyed) {
+        req[getAbortController]().abort()
+      }
+    })
 
     try {
       res = fetchCallback(req, { incoming, outgoing } as HttpBindings) as
--- a/test/listener.test.ts
+++ b/test/listener.test.ts
@@ -89,3 +89,119 @@ describe('Error handling - async fetchCallback', () => {
     expect(res.text).toBe('error handler did not return a response')
   })
 })
+
+describe('Abort request', () => {
+  let onAbort: (req: Request) => void
+  let reqReadyResolve: () => void
+  let reqReadyPromise: Promise<void>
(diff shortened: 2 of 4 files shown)


src/listener.ts
```diff
@@ -143,6 +143,13 @@ export const getRequestListener = (
     // so generate a pseudo Request object with only the minimum required information.
     const req = newRequest(incoming)
 
+    // Detect if request was aborted.
+    incoming.socket.on('close', () => {
+      if (incoming.aborted) {
```
M4RC3L05, line 148: From the nodejs docs it appears that the `aborted` property is deprecated, and as a replacement, we should check the `destroyed` (https://nodejs.org/dist/latest-v20.x/docs/api/http.html#messageaborted).

For some reason if i use it, it makes most of the tests fail, do not know if this is something related with supertest...

test/listener.test.ts
```diff
@@ -89,3 +89,42 @@ describe('Error handling - async fetchCallback', () => {
     expect(res.text).toBe('error handler did not return a response')
   })
 })
+
+describe('Abort request', () => {
```
M4RC3L05, line 93: Is there any more tests that i should include?

src/listener.ts
M4RC3L05: Resolved, had to listen to the close event on the response and check the `incoming.socket.destroyed`

yusukebe:
Hi @M4RC3L05 

Thank you for the PR. Looks good to me.

@usualoma Could you too review this?

usualoma:
I'm preparing an improvement plan now, so wait a minute.

usualoma:
Hi @M4RC3L05 
Thank you. The implementation policy makes sense to me!

I hope to achieve the following two points as well.

* The `aborted` property would be `true`.
　 * https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/aborted
* not use `setInterval()` in tests.

As an example, I would like to have something like the following PR, can you incorporate it into your PR?
https://github.com/honojs/node-server/pull/142

M4RC3L05:
Hi, sure, I can work on making does changes

M4RC3L05:
Updated with the improvements pointed out on https://github.com/honojs/node-server/pull/142 (thx btw :pray:).

Also, added some more tests, please @usualoma @yusukebe let me know if they are relevant or not.

test/request.test.ts
```diff
@@ -40,6 +40,34 @@ describe('Request', () => {
       expect(req).toBeInstanceOf(global.Request)
       expect(req.url).toBe('http://localhost/foo.txt')
     })
+
+    it('should generate only one `AbortController` per `Request` object created', async () => {
```
M4RC3L05, line 44: Just a sanity check to make sure we do not create a new AbortController every time we access it, or any other property since it calls `getRequestCache`.

usualoma:
@M4RC3L05 Thank you. Looks good!

yusukebe:
@M4RC3L05 @usualoma Thanks both!

I'd like to merge, as @usualoma as co-author.

@M4RC3L05 
One thing, could run `yarn lint:fix && yarn format:fix` to fix the CI error?

M4RC3L05:
Done

yusukebe:
@M4RC3L05 Thanks!!

PR #145, opened 2024-02-17
feat: use internal body if available for returning the response in its original form as much as possible

### LINKED ISSUE #144: should use `writeFromReadableStream()` for a `ReadableStream`
It's time to refactor response from an un-cached request object

https://github.com/usualoma/node-server/blob/b34b9a94c8006810fd21c5189cb663982fa98e59/src/listener.ts#L86-L130

Although it is partly correct to base the decision on the resHeaderRecord, it is more expected to base the decision on the type of the body.

Currently, if the body is not "Transfer-Encoding: chunked", it is returned from arrayBuffer even if the original body is a readbleStream, but we should use `writeFromReadableStream()` for this.

usualoma:
Fixes #144

Reuse the following implementation idea to get the body inside the nodejs Response object and use it to return a response if available.

https://github.com/honojs/node-server/blob/a8117ce9e4f6b1a2615dde1d1b4c2428b801cfa2/src/utils.ts#L13-L24

### DIFF
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -72,8 +72,7 @@ const responseViaResponseObject = async (
     }
   }
 
-  const isCached = cacheKey in res
-  if (isCached) {
+  if (cacheKey in res) {
     return responseViaCache(res as Response, outgoing)
   }
 
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -72,13 +72,9 @@ const responseViaResponseObject = async (
     }
   }
 
-  try {
-    const isCached = cacheKey in res
-    if (isCached) {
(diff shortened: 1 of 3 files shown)


usualoma:
By the way, the following code is still used.

https://github.com/honojs/node-server/blob/a25fa2a2c558f575ef0bf35fcd9faf055c47d2e2/src/listener.ts#L96-L130

Removing this will cause the following tests to fail

https://github.com/honojs/node-server/blob/a25fa2a2c558f575ef0bf35fcd9faf055c47d2e2/test/server.test.ts#L78-L83

yusukebe (approved):
Great! Looks good to me!

Issue #147, opened 2024-02-18
`app.mount()` is not working with `@hono/node-server`


yusukebe:
This code does not work well:

```ts
import { serve } from '@hono/node-server'
import { Hono } from 'hono'

const app = new Hono()

app.mount('/mount', () => {
  return new Response('foo')
})

serve(app)
```

The error message:

```
TypeError: Cannot read properties of undefined (reading 'keepalive')
    at get keepalive [as keepalive] (node:internal/deps/undici/undici:5612:29)
    at Object.RequestInit (node:internal/deps/undici/undici:1779:33)
    at new Request (node:internal/deps/undici/undici:5253:34)
    at Request (/Users/yusuke/work/node/my-app-node/node_modules/@hono/node-server/dist/index.js:55:5)
    at Array.handler (/Users/yusuke/work/node/my-app-node/node_modules/hono/dist/cjs/hono-base.js:191:9)
    at Hono.dispatch (/Users/yusuke/work/node/my-app-node/node_modules/hono/dist/cjs/hono-base.js:235:37)
    at fetch (/Users/yusuke/work/node/my-app-node/node_modules/hono/dist/cjs/hono-base.js:82:19)
    at Server.<anonymous> (/Users/yusuke/work/node/my-app-node/node_modules/@hono/node-server/dist/index.js:399:13)
    at Server.emit (node:events:514:28)
    at Server.emit (node:domain:488:12)
```

yusukebe:
Hi @usualoma 

Can you see this issue? I think this is a definite bug and should be fixed.

usualoma:
Hi @yusukebe 

I think the following PR can fix this issue.
https://github.com/honojs/node-server/pull/148

ManuelFernando:
@yusukebe I had same problem because using app.mount() I can't access to any headers. These are protected, is there a way to read the request. For example when I use **req.headers['content-type']** -> this return **undefined**

yusukebe:
Hi @ManuelFernando 

I can't reproduce it. Please provide a minimal project to reproduce it.

ManuelFernando:
> Hi @ManuelFernando
> 
> I can't reproduce it. Please provide a minimal project to reproduce it.

when it's used console.log(req.headers) in the console appears string representation but any of their field can be accessed

```
import { serve } from '@hono/node-server';
import { Hono } from 'hono';

const app = new Hono();

app.mount('/status', (req) => {
  console.log(req.headers.accept); // -> return undefined
[...]

PR #148, opened 2024-02-18
feat: support keepalive property for Request object

usualoma:
fixes https://github.com/honojs/node-server/issues/147


yusukebe:
@usualoma 

Thanks! Works well. Now I can do what I want to do!

PR #153, opened 2024-03-22
feat: ignore response if already sent by raw level outgoing API.

usualoma:


yusukebe (approved):
LGTM!

yusukebe:
Hi @usualoma 

Thanks! This is a good and simple solution. Let's go with this.

Issue #154, opened 2024-03-25
Optimisations breaking `instanceof Response` checks


ezg27:
Hi there! It looks as though `@hono/node-server` overriding Web API globals is causing valid `instanceof Response` checks to fail in Node.js when they succeed in other applications, including the same Hono apps in other runtimes e.g. Bun, Deno, Cloudflare.

It appears to be related to the changes made in the following PR to reduce the overhead of `Request` and `Response` instance creation by replacing them with pseudo objects: https://github.com/honojs/node-server/pull/95.

Since those changes were released, attempting to do `instanceof Response` checks against a response returned from native `fetch` in Node.js will return `false` (pre v1.3.0 this would correctly return `true`). Just importing `@hono/node-server` is enough for the globals to be overridden, so the following is all that's needed to reproduce the issue (using Node v20):

```js
import '@hono/node-server';

const response = await fetch('https://www.example.com');
console.log(response instanceof Response); // false
```

While i understand the motivation to try and reduce overhead in Node compared to other runtimes, I'm not sure if abstractions like this are the way to go when their implementations can leak into application code. However, i'm also not sure it'll be possible to achieve similar microbenchmark results as those reported in the above PR while still using the the native `Request` and `Response`.

For now i can work around the above issues by using `patch-package` to remove the imports/optimisations from `./request.js` and `./response.js` so that `instanceof` checks work as expected. However, if the aim is to keep the current overrides, would you be open to adding an option to disable them as part of the library (e.g. an optional `disableOverrides` that can be passed to `serve`)? I'd be happy to raise a PR for this, but will defer to you on the best way forward. At the very least i reckon it's worth documenting the overrides and these potential issues (the README currently still says `It utilizes web standard APIs implemented in Node.js version 18 or higher.`)

rafaell-lycan:
I believe this is a great opportunity to also make sure testing regression for cases like this one.

yusukebe:
Hi @ezg27 !

Sorry for the super delayed reply. This is an important issue, and something must be done about it.

> However, if the aim is to keep the current overrides, would you be open to adding an option to disable them as part of the library (e.g. an optional `disableOverrides` that can be passed to `serve`)?

This is good. I don't know if the name `disableOverrides` is appropriate, but we should have an option not to use the Lightweight version of the pseudo object. When it is implemented, it will solve the problems commented [here](
https://github.com/honojs/vite-plugins/issues/107#issuecomment-2037991839).

@usualoma Do you have any thoughts?

[...]

PR #156, opened 2024-04-08
feat: enable to skip overriding global Request and Response classes

usualoma:


usualoma:
@yusukebe 
Sorry for the delay - the test didn't go through very well and it took a long time.
The code that is executed has not been changed from https://github.com/honojs/node-server/issues/154#issuecomment-2041659272 so I don't think it is a problem, but please review it just in case.

rafaell-lycan (approved):
LGTM

yusukebe (approved):
LGTM!

yusukebe:
Hi @usualoma 

The test matter is not a problem. Great! I'll merge now. Thanks!

PR #162, opened 2024-04-19
fix: catch ERR_INVALID_URL error in listener

### LINKED ISSUE #159: Denial of Service in newRequest
Hi, Denial of Service vulnerability discovered.

## PoC

Source

```javascript
import { Hono } from 'hono'

const app = new Hono()

app.get('/*', (c) => {
    return c.text("hello")
})

serve(app, (info) => {
    console.log(`Listening on http://localhost:${info.port}`) // Listening on http://localhost:3000
})
```

Request

```bash
curl localhost:3000/ -H "Host: "
```

As a result, the application stops.

> Exception has occurred: TypeError: Invalid URL

In this case, the argument passed to `new URL` will be `http:///` .

https://github.com/honojs/node-server/blob/c22f7500dc64bc08e1f8321f81d148eb061ec256/src/request.ts#L117-L126

## **Suggested fix**

It is recommended to validate the Host header.

---

In addition, it is possible to manipulate the `c.req.path` by including `/` in the Host header.
This has not so far led to any direct damage, but it could be used for WAF bypassing, etc., so it is recommended to add this to the validation process.

Source

```jsx
import { Hono } from 'hono'

const app = new Hono()

app.get('/*', (c) => {
    return c.json({
        c_env_incoming_url: c.env.incoming.url,
        c_req_url: c.req.url,
        c_req_path: c.req.path,
    })
})

serve(app, (info) => {
    console.log(`Listening on http://localhost:${info.port}`) // Listening on http://localhost:3000
})
```

Request

```bash
curl localhost:3000/aaa -H "Host: example.com/bbb?" 
```

Response

```json
{
	"c_env_incoming_url":"/aaa",
	"c_req_url":"http://example.com/bbb?/aaa",
	"c_req_path":"/bbb"
}
```

usualoma:
fixes #159

### DIFF
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -159,18 +159,18 @@ export const getRequestListener = (
   ) => {
     let res
 
-    // `fetchCallback()` requests a Request object, but global.Request is expensive to generate,
-    // so generate a pseudo Request object with only the minimum required information.
-    const req = newRequest(incoming)
-
-    // Detect if request was aborted.
-    outgoing.on('close', () => {
-      if (incoming.destroyed) {
-        req[getAbortController]().abort()
-      }
-    })
-
     try {
+      // `fetchCallback()` requests a Request object, but global.Request is expensive to generate,
+      // so generate a pseudo Request object with only the minimum required information.
+      const req = newRequest(incoming)
+
+      // Detect if request was aborted.
+      outgoing.on('close', () => {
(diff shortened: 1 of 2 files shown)


yusukebe (approved):

yusukebe:
@usualoma 

Thanks for the quick fix! I'll merge it right now.

Issue #161, opened 2024-04-19
Return 400 or 500 if the Host header contains a path.


usualoma:
Currently, the path is passed directly to the URL if it is included in the host header, but should return 400.

app
```
const app = new Hono()
app.get('/', (c) => c.text(c.req.url))
```

request
```
$ curl http://localhost:3002 -H 'Host: localhsot/?key=value'
http://localhsot/?key=value/
```

### How does it behave on other servers?

`bun` and `deno` have the same result as the current `node-server`.
`wrangler dev` results in a 500 error

yusukebe:
This was closed by #163. Thanks!

PR #163, opened 2024-04-19
feat: validate incoming host header

### LINKED ISSUE #160: Determine behaviour for HTTP1.0 request w/o Host header
The following results are currently returned

app
```ts
const app = new Hono()
app.get('/', (c) => c.text(c.req.raw.url))

const port = 3002
serve({
    fetch: app.fetch,
    port,
})
```

request
```
$ curl http://127.0.0.1:3002 -H 'Host:' --http1.0
http://undefined/
```

HTTP1.0 request w/o Host header is not an illegal request and should not be an error, but http://undefined/ is an incorrect URL.

Nevertheless, even reverse proxies that speak HTTP1.0 would normally add a Host header, so it may be acceptable to say "Request without Host header is an error”.

If the aim is to support all cases, it may be possible to specify the default hostname as an option, but this seems a bit excessive.
```diff
diff --git a/src/listener.ts b/src/listener.ts
index 82655d2..17855b3 100644
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -142,6 +142,7 @@ export const getRequestListener = (
   options: {
     errorHandler?: CustomErrorHandler
     overrideGlobalObjects?: boolean
+    defaultHostnames?: string
   } = {}
 ) => {
   if (options.overrideGlobalObjects !== false && global.Request !== LightweightRequest) {
```

usualoma:
fixes #160, #161

### What is this?

Validate the content of the host header of the request for security purposes.

### Invalid values

If the following values are specified in the host header, a "400 Bad Request" is returned.

* Empty string (if no default hostname is specified)
* Including path or query parameter
    * `localhost/path` `localhost?key=value`
* Including invalid charactors
    * `localhost<script>`

### Valid values

Host headers are only validated by character type, so "hostnames with invalid structure" or "hostnames that cannot be looked up" are considered valid and do not result in a "400 Bad Request". For example, the following host headers are considered valid
* `.` `www...example.com`  `()`

### Default hostname

If you also want to accept HTTP 1.0 requests w/o Host header, you can call `serve()` as follows to ensure that the default hostname is used without error if there is no host header (or an empty string is specified).

```ts
serve({
    fetch: app.fetch,
    hostname: 'localhost', // should be associated with the IP address of the server
})
```

### Tasks

* [x] implement
* [x] test

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,3 +1,4 @@
 export { serve, createAdaptorServer } from './server'
 export { getRequestListener } from './listener'
+export { RequestError } from './request'
 export type { HttpBindings, Http2Bindings } from './types'
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -1,6 +1,11 @@
 import type { IncomingMessage, ServerResponse, OutgoingHttpHeaders } from 'node:http'
 import type { Http2ServerRequest, Http2ServerResponse } from 'node:http2'
-import { getAbortController, newRequest, Request as LightweightRequest } from './request'
+import {
+  getAbortController,
+  newRequest,
+  Request as LightweightRequest,
+  toRequestError,
+} from './request'
 import { cacheKey, getInternalBody, Response as LightweightResponse } from './response'
 import type { CustomErrorHandler, FetchCallback, HttpBindings } from './types'
 import { writeFromReadableStream, buildOutgoingHttpHeaders } from './utils'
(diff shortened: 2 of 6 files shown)


yusukebe (approved):

yusukebe:
Hi @usualoma 

Thank you for handling it. I'll merge this and release the new version right now. The most important thing is to release it first.

usualoma:
@yusukebe Thank you!

PR #164, opened 2024-04-22
Update Request to work with ReadableStream polyfills

nicksrandall:
Some web frameworks (like Remix) inject polyfills for ReadableStream into the Node Runtime.  This change makes sure that the server will work correctly with those polyfills. Otherwise the `instanceof` check would fail.

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -31,7 +31,8 @@ export class Request extends GlobalRequest {
     if (typeof input === 'object' && getRequestCache in input) {
       input = (input as any)[getRequestCache]()
     }
-    if (options?.body instanceof ReadableStream) {
+    // Check if body is ReadableStream like. This makes it compatbile with ReadableStream polyfills. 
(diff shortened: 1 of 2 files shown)


nicksrandall:
More info on Remix Polyfills: 
- https://remix.run/docs/en/main/other-api/node
- https://github.com/remix-run/remix/blob/main/packages/remix-node/globals.ts#L59

I borrowed this check from the polyfill implementation here:
- https://github.com/MattiasBuelens/web-streams-polyfill/blob/master/src/lib/readable-stream/readable-stream-like.ts#L30-L32

yusukebe:
Hi @nicksrandall !

Thanks for the PR. It looks good to me. One thing, could you run `yarn format:fix && yarn lint:fix` to fix the format?

nicksrandall:
@yusukebe Done

yusukebe:
@nicksrandall Thanks! Merging now.

PR #170, opened 2024-05-21
ci: include tests for node 22 version

Jayllyz:
Node.js 22 will enter long-term support (LTS) in October, so I think it would be a good idea to start running CI on it now.

I've also added workflow permissions and increased action versions.


yusukebe:
@Jayllyz 

Thanks! Adding the Node 22 version is good! Merging now.

PR #172, opened 2024-05-29
fix:  memory leak by AbortController

### LINKED ISSUE #171: Memory leak with `hono`, `@hono/node-server`, `@hono/graphql-server`
It seems like there is a memory leak when using `hono`, `@hono/node-server`, and `@hono/graphql-server` together.

Minimal reproduction mostly from the example in `@hono/graphql-server`:

```js
import { Hono } from "hono";
import { graphqlServer } from "@hono/graphql-server";
import { buildSchema } from "graphql";

setInterval(() => {
  const before = process.memoryUsage().heapUsed / 1024 / 1024;
  global.gc?.();
  const after = process.memoryUsage().heapUsed / 1024 / 1024;
  console.log({ before: before.toFixed(2), after: after.toFixed(2) });
}, 1_000);

export const app = new Hono();

const schema = buildSchema(`
type Query {
  hello: String
}
`);

const rootResolver = (ctx) => {
  return {
    hello: () => "Hello Hono!",
  };
};

app.use(
  "/graphql",
  graphqlServer({
    schema,
    rootResolver,
  })
);

serve(app);
```

Result of running the server and then sending requests with autocannon after a few seconds:

```sh
autocannon -c 100 -d 10 -m POST -H "Content-Type: application/json" -b '{"query":"{ hello }"}' http://localhost:3000/graphql
```

```sh
(base) kylescott@Kyles-MacBook-Pro-2 hono-node-repro % node --expose-gc index.js
{ before: '8.86', after: '6.72' }
{ before: '6.83', after: '6.75' }
{ before: '6.77', after: '6.75' }
{ before: '6.77', after: '6.75' }
{ before: '6.82', after: '6.76' }
{ before: '6.78', after: '6.76' }
{ before: '6.83', after: '6.77' }
{ before: '55.64', after: '37.63' }
{ before: '153.87', after: '97.79' }
{ before: '205.73', after: '152.70' }
{ before: '262.65', after: '208.79' }
{ before: '310.95', after: '259.03' }
{ before: '355.99', after: '308.32' }
{ before: '400.77', after: '354.94' }
{ before: '444.92', after: '398.86' }
{ before: '489.06', after: '443.80' }
{ before: '523.01', after: '484.06' }
{ before: '565.09', after: '523.40' }
```

https://github.com/kyscott18/hono-node-mem-leak

usualoma:
Fixes #171, https://github.com/honojs/hono/issues/2816

Two issues needed to be corrected in this issue.

### `incoming.destroyed` is true for all POST requests

[The documentation](https://nodejs.org/api/http.html#requestaborted) says "Deprecated. check request.destroyed instead."

CleanShot 2024-05-30 at 05 17 02@2x

However, incoming was flagged for “autodestroy" and `incoming.destroy()` was always called when the request body was read to the end. In other words, `.abort()` was always called in POST requests. POST requests complete successfully, so there are no major problems, but I think it was not a good behaviour.

https://github.com/nodejs/node/blob/c0c598d753d5ce7e1721215c6e6254dc02c023a8/lib/internal/streams/readable.js#L1690-L1716

As [errored](https://nodejs.org/api/stream.html#readableerrored) seemed more suitable for our application, I decided to use it.

### `.abort()` (call without arguments) causes memory leak

The real problem of memory leaks is here, which was an interesting situation.

Calling `abort()` with no arguments uses an instance of `DOMException`, but (presumably) this object is referenced including its children and grandchildren, so it appears that a cycle of references occurs and a instance of AbortController cannot be freed.

https://github.com/nodejs/node/blob/c0c598d753d5ce7e1721215c6e6254dc02c023a8/lib/internal/abort_controller.js#L395-L397
https://github.com/nodejs/node/blob/c0c598d753d5ce7e1721215c6e6254dc02c023a8/lib/internal/abort_controller.js#L368-L378

For "reason", not only for `DOMException`, it seems to be necessary to pass a string, as specifying an `Object` seems to cause a memory leak.

### tsconfig.json

In order to use `FinalisationRegistry` in server.test.ts, the settings in tsconfig.json needed to be reflected in server.test.ts, so tsconfig.json was changed.

32c09a9b1b56220c8c3bdd1b0c01189933d41ecc

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     }
   },
   "scripts": {
-    "test": "jest",
+    "test": "node --expose-gc ./node_modules/.bin/jest",
     "build": "tsup",
     "watch": "tsup --watch",
     "postbuild": "publint",
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -179,7 +179,7 @@ export const getRequestListener = (
       // Detect if request was aborted.
       outgoing.on('close', () => {
         if (incoming.errored) {
-          req[getAbortController]().abort()
+          req[getAbortController]().abort(incoming.errored.toString())
         }
       })
 
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -178,7 +178,7 @@ export const getRequestListener = (
 
       // Detect if request was aborted.
       outgoing.on('close', () => {
-        if (incoming.destroyed) {
+        if (incoming.errored) {
           req[getAbortController]().abort()
         }
       })
--- a/test/server.test.ts
+++ b/test/server.test.ts
@@ -8,7 +8,7 @@ import { compress } from 'hono/compress'
 import { poweredBy } from 'hono/powered-by'
 import { stream } from 'hono/streaming'
 import request from 'supertest'
-import { GlobalRequest, Request as LightweightRequest } from '../src/request'
+import { GlobalRequest, Request as LightweightRequest, getAbortController } from '../src/request'
 import { GlobalResponse, Response as LightweightResponse } from '../src/response'
 import { createAdaptorServer } from '../src/server'
 import type { HttpBindings } from '../src/types'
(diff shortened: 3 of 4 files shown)


usualoma:
@yusukebe Would you please review?

usualoma:
`node --expose-gc` is required for executing `global.gc?.()`

kyscott18:
> `node --expose-gc` is required for executing `global.gc?.()`

yes

kyscott18:
This branch fixed my original bug! Thank you!

yusukebe (approved):
LGTM!

yusukebe:
Hi @usualoma !

Thank you for the fix and the test is great! I'll merge this now.

PR #175, opened 2024-06-13
fix: avoid error when using TRACE method

### LINKED ISSUE #167: HTTP TRACE method crashes server
```
curl -k -v -X TRACE http://localhost:8989
```

crashes server with:

```
TypeError: 'TRACE' HTTP method is unsupported.
    at new Request (node:internal/deps/undici/undici:7245:19)
    at new Request (./node_modules/@hono/node-server/dist/index.js:55:5)
```

usualoma:
fixes #167

The Request object defined in undici is an error if we try to create it by specifying the TRACE method. In this pull request, create it by specifying the GET method to avoid this.

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -62,6 +62,17 @@ const newRequestFromIncoming = (
     signal: abortController.signal,
   } as RequestInit
 
+  if (method === 'TRACE') {
+    init.method = 'GET'
+    const req = new Request(url, init)
+    Object.defineProperty(req, 'method', {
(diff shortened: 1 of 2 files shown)


yusukebe (commented):
LGTM!

yusukebe (approved):

yusukebe:
> The Request object defined in undici is an error if we try to create it by specifying the TRACE method.

Ah, I see! Thanks!

PR #188, opened 2024-08-18
fix: return response from res.body if internal data is not ready to be returned directly

### LINKED ISSUE #186: Cannot return `Buffer` when ETag middleware is used
When running the code

```js
import { Hono } from 'hono'
import { etag } from 'hono/etag'

const app = new Hono()
app.use(etag())
app.get('/', (c) => c.body(Buffer.from('Hello Hono!')))
serve({ fetch: app.fetch })
```

and running `curl localhost:3000`, the server crashes:

```
node:internal/buffer:961
    super(bufferOrLength, byteOffset, length);
    ^

TypeError: Cannot perform Construct on a detached ArrayBuffer
    at new Uint8Array (<anonymous>)
    at new FastBuffer (node:internal/buffer:961:5)
    at Function._uint8ArrayToBuffer (node:stream:144:10)
    at _write (node:internal/streams/writable:477:22)
    at Writable.write (node:internal/streams/writable:508:10)
    at ServerResponse._writeRaw (node:_http_outgoing:443:17)
    at ServerResponse._send (node:_http_outgoing:420:15)
    at write_ (node:_http_outgoing:1018:15)
    at ServerResponse.end (node:_http_outgoing:1102:5)
    at responseViaResponseObject (/Users/mruder/git-repos/hono-buffer/node_modules/@hono/node-server/dist/index.js:392:16)
```

usualoma:
fixes #186

"etag" middleware does `res.clone()`, but if you use `Buffer.from`, the internal data is shared, so that once data is read with ’etag’, it could not be re-read on the server side.

https://github.com/usualoma/hono/blob/915a8234975b8b875edcdf6a3acc8280dfc898af/src/middleware/etag/index.ts#L66

In such cases, if you read from `res.body`, the undici handles it well, so you can fallback to `res.body`.

### DIFF
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -92,18 +92,27 @@ const responseViaResponseObject = async (
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const internalBody = getInternalBody(res as any)
   if (internalBody) {
-    if (internalBody.length) {
-      resHeaderRecord['content-length'] = internalBody.length
-    }
-    outgoing.writeHead(res.status, resHeaderRecord)
-    if (typeof internalBody.source === 'string' || internalBody.source instanceof Uint8Array) {
-      outgoing.end(internalBody.source)
-    } else if (internalBody.source instanceof Blob) {
-      outgoing.end(new Uint8Array(await internalBody.source.arrayBuffer()))
+    const { length, source, stream } = internalBody
+    if (source instanceof Uint8Array && source.byteLength !== length) {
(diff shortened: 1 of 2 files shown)


usualoma:
@yusukebe Would you please review?

yusukebe (approved):
LGTM!

yusukebe:
@usualoma 

Thank you! Good fix. I'll merge it and release a new version.

PR #192, opened 2024-08-26
Declare `hono` as peer dependency

### LINKED ISSUE #191: `hono` is not declared as peer dependency
Given a Hono project under Node.js using the Yarn package manager with Yarn PnP enabled, when attempting to import `@hono/node-server`, then the import fails with the message

```
  The Yarn Plug'n'Play manifest forbids importing "hono" here because it's not listed as a
  dependency of this package:
```

This is because `@hono/node-server` uses imports from `hono` (e.g. [here](https://github.com/honojs/node-server/blob/e42500f07d1eaff989cb63710afb8737f7b3c055/src/serve-static.ts#L4)) without declaring it as either a dependency or a peer dependency.

A possible workaround at the moment is to set

```yaml
packageExtensions: 
  "@hono/node-server@*": 
    peerDependencies: 
      "hono": "*"
```

in your `.yarnrc.yml`

marvinruder:
* Fixes #191

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,5 +95,8 @@
     "ts-jest": "^29.1.1",
     "tsup": "^7.2.0",
     "typescript": "^5.3.2"
+  },
+  "peerDependencies": {
+    "hono": "^4"
   }
 }


yusukebe (approved):
LGTM!

PR #197, opened 2024-09-12
chore(lint): use eslint v9

Jayllyz:
Following https://github.com/honojs/hono/pull/3393

Removed tests on Node 19 because it's not a maintained version anymore.
https://nodejs.org/en/about/previous-releases


yusukebe:
@Jayllyz 

Cool! I'll merge it now. Thank you!

PR #199, opened 2024-09-15
feat(serve-static): add `precompressed` option

yusukebe:
Implemented the same feature of https://github.com/honojs/hono/pull/3366

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,7 +199,7 @@ app.use(
 You can specify handling when the requested file is found with `onFound`.
 
 ```ts
-app.get(
+app.use(
   '/static/*',
   serveStatic({
     // ...
@@ -222,6 +222,19 @@ app.use(
     onNotFound: (path, c) => {
       console.log(`${path} is not found, request to ${c.req.path}`)
     },
+  })
+)
+```
+
+#### `precompressed`
+
+The `precompressed` option checks if files with extensions like `.br` or `.gz` are available and serves them based on the `Accept-Encoding` header. It prioritizes Brotli, then Zstd, and Gzip. If none are available, it serves the original file.
+
+```ts
+app.use(
+  '/static/*',
+  serveStatic({
+    precompressed: true,
   })
 )
 ```
--- a/src/serve-static.ts
+++ b/src/serve-static.ts
@@ -11,10 +11,17 @@ export type ServeStaticOptions<E extends Env = Env> = {
   root?: string
   path?: string
   index?: string // default is 'index.html'
+  precompressed?: boolean
   rewriteRequestPath?: (path: string) => string
   onFound?: (path: string, c: Context<E>) => void | Promise<void>
   onNotFound?: (path: string, c: Context<E>) => void | Promise<void>
 }
+
+const ENCODINGS = {
+  br: '.br',
+  zstd: '.zst',
+  gzip: '.gz',
+} as const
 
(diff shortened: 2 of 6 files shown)


yusukebe:
Hey @usualoma !

Can your review this?

README.md
```diff
@@ -199,7 +199,7 @@ app.use(
 You can specify handling when the requested file is found with `onFound`.
 
 ```ts
-app.get(
+app.use(
```
yusukebe, line 202: Other examples are using `use`.

usualoma:
I'll check it out, just give me a minute.

usualoma:
Sorry I haven't commented on this at https://github.com/honojs/hono/pull/3366. I think it is a very good feature, but the implementation could use some improvement in my opinion.

I have created https://github.com/honojs/node-server/pull/200 and would like you to consider merging

### calling `sort` or `Object.keys`

It would be better to avoid calling `sort` or `Object.keys` on every request.

In my environment, I got the following results

```ts
import { run, bench, group } from "mitata";

const ENCODINGS = {
  br: ".br",
  zstd: ".zst",
  gzip: ".gz",
} as const;
const ENCODINGS_ORDERED_KEYS = Object.keys(ENCODINGS) as (keyof typeof ENCODINGS)[]

const filterAndSort = (ae: string) => {
  const acceptEncodings =
    ae
      .split(",")
      .map((encoding) => encoding.trim())
      .filter((encoding): encoding is keyof typeof ENCODINGS =>
        Object.hasOwn(ENCODINGS, encoding)
      )
      .sort(
        (a, b) =>
          Object.keys(ENCODINGS).indexOf(a) - Object.keys(ENCODINGS).indexOf(b)
      ) ?? [];

  for (const encoding of acceptEncodings) {
    break;
  }
};

const set = (ae: string) => {
  const acceptEncodingsSet = new Set(ae.split(",").map((encoding) => encoding.trim()));

  for (const encoding of ENCODINGS_ORDERED_KEYS) {
    if (!acceptEncodingsSet.has(encoding)) {
      continue;
    }
    break;
  }
};

group("filterAndSort vs set", () => {
  bench("filterAndSort", () => filterAndSort("br,gzip,zstd"));
  bench("set", () => set("br,gzip,zstd"));
});

run();
```

```
cpu: Apple M2 Pro
runtime: node v20.14.0 (arm64-darwin)

benchmark          time (avg)             (min … max)       p75       p99      p999
----------------------------------------------------- -----------------------------
• filterAndSort vs set
----------------------------------------------------- -----------------------------
filterAndSort     284 ns/iter       (258 ns … 562 ns)    299 ns    357 ns    498 ns
set               127 ns/iter       (118 ns … 294 ns)    126 ns    178 ns    236 ns

summary for filterAndSort vs set
  set
   2.24x faster than filterAndSort
```


### Find files only if mime type is compressible

If all that is delivered from static is css or js, it would be a useless check, but image files are often placed in static and may be files that are not compressed additionally. Access to the file system is generally very heavy, so it would be better to check the mime type and look for it only when the file should be compressed.

yusukebe:
@usualoma Thanks! I'll check it later.

yusukebe:
Hi @usualoma 

Fully agree with your improvements. Using `Set` is fast, and choosing a compressible file is good! I'll merge your branch into this. Thank you!

PR #201, opened 2024-09-21
fix(serve-static): use application/octet-stream if the mime type is not detected

usualoma:
This pull request is a port of the following pull request.

https://github.com/honojs/hono/pull/3415


yusukebe:
@usualoma 

Thanks! Merging.

PR #202, opened 2024-09-30
feat(serve-static): support absolute path for `root`

### LINKED ISSUE #187: add support for absolute path to serveStatic root
Could you add support for _absolute path_ for the root of the serveStatic?
In my use-case [designix-uis](https://www.npmjs.com/package/designix-uis), an absolute path to the folder containing the static files is required as tool can be called from any directory. Currently i"m using Express, but i"d like to replace it with Hono and this feature is missing.
The following patch might be sufficient for supporting the absolute path:
```
-      path = addCurrentDirPrefix(path)
+      if (!/^\//.test(path)) {
+        path = addCurrentDirPrefix(path)
+      }
```

yusukebe:
This PR enables the serve static to support an absolute path for the `root` option.

```ts
import { serveStatic } from '@hono/node-server'
import { serve } from '@hono/node-server/serve-static'
import { Hono } from 'hono'

const app = new Hono()
app.use('/static/*', serveStatic({ root: '/home/yusuke/app' }))

serve(app)
```

Closes #187
Related to https://github.com/honojs/hono/pull/3420

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ import { serveStatic } from '@hono/node-server/serve-static'
 app.use('/static/*', serveStatic({ root: './' }))
 ```
 
-Note that `root` must be _relative_ to the current working directory from which the app was started. Absolute paths are not supported.
+Note that `root` must be _relative_ to the current working directory from which the app was started.
 
 This can cause confusion when running your application locally.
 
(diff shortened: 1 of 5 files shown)


yusukebe:
Hey @usualoma !

Can you review this?

yusukebe:
@usualoma 

I have some thoughts related to https://github.com/honojs/node-server/issues/203 and we don't have to hurry to merge this because this PR may be not needed. So, please don't review this now. Thank you!

PR #206, opened 2024-10-09
fix(serve): support ipv6 by default

sinasab:
For node's server.listen, if no hostname param is specified it'll adaptively pick the unspecified ipv6 address (`::`) when available or the unspecified ipv4 address (`0.0.0.0`) otherwise. This fix gets rid of the behavior where we manually default to `0.0.0.0` and rely on node's underlying handling of this.

Since most modern systems should support ipv6, I also added a test case on this new behavior.

See here for the docs on node: https://nodejs.org/api/net.html#serverlistenport-host-backlog-callback:~:text=If%20host%20is%20omitted%2C%20the%20server%20will%20accept%20connections%20on%20the%20unspecified%20IPv6%20address%20(%3A%3A)%20when%20IPv6%20is%20available%2C%20or%20the%20unspecified%20IPv4%20address%20(0.0.0.0)%20otherwise.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -21,7 +21,7 @@ export const serve = (
   listeningListener?: (info: AddressInfo) => void
 ): ServerType => {
   const server = createAdaptorServer(options)
-  server.listen(options?.port ?? 3000, options.hostname ?? '0.0.0.0', () => {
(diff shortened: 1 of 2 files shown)


yusukebe (commented):
LGTM!

PR #208, opened 2024-10-25
fix(serve-static): add error handler for decoding uri components

alumowa:
This PR adds an error handler for potentially thrown `URIError` during `decodeURIComponent` that should result in a HTTP 404 rather than an HTTP 500 error. 

This may happen if the encoded uri component contains an [invalid sequence encoding](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent):

> [URIError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/URIError)
Thrown if encodedURI contains a % not followed by two hexadecimal digits, or if the escape sequence does not encode a valid UTF-8 character.

Also added a test for this case 👓

### DIFF
--- a/src/serve-static.ts
+++ b/src/serve-static.ts
@@ -63,7 +63,14 @@ export const serveStatic = (options: ServeStaticOptions = { root: '' }): Middlew
       return next()
     }
 
-    const filename = options.path ?? decodeURIComponent(c.req.path)
+    let filename: string
+
+    try {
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
Hi @alumowa 

Thank you for the PR. I also think it's okay to return a `404`. I'll merge it and release a new version.

PR #211, opened 2024-10-31
fix: Object.entries is called with a null value

mjad218:
A small change to the same function. 
I am using hono with Nest.js and somehow `buildOutgoingHttpHeaders` is called with a `null` argument.

I have updated the function to address this edge case.


yusukebe:
Hi @mjad218 

The PR https://github.com/honojs/node-server/pull/212 is merged. So, is this change unnecessary?

mjad218:
@yusukebe 
No this should be closed.

PR #216, opened 2024-11-09
fix: conninfo returns server IP

### LINKED ISSUE #215: Is the behavior of getConnInfo as expected ?
In the current implementation, the remote.address field is set with the value obtained from `const address = bindings.incoming.socket.address() as AddressInfo`. However, this method, `socket.address()`, retrieves the server's socket address rather than the remote client’s address.

https://github.com/honojs/node-server/blob/350f9b7fba008e26231771e3d82087250e6c9561/src/conninfo.ts#L13

This raises two potential issues:

Field Name and Value Mismatch
The field remote.address typically implies it would contain the address of the remote client (i.e., the origin of the connection). Using `socket.address()` may lead to misunderstandings, as it actually refers to the server’s address.

Expected Address Information
If the goal is to store the remote client's address, we may need to use something like `socket.remoteAddress` instead, which would reflect the client's IP.

Background:

I noticed this issue while using Hono’s [ipRestriction](https://hono.dev/docs/middleware/builtin/ip-restriction) middleware in a Node.js environment. The documentation suggested using ipRestriction(getConnInfo, {}), but it didn’t work as expected.

If getConnInfo itself is behaving as intended, then there may be a need to adjust the ipRestriction middleware to align with expected functionality.

I haven’t tested this in other runtimes, but they might have the same issue.

Apologies if I misunderstood the intended behavior.

nakasyou:
Fixes #215.

### DIFF
--- a/src/conninfo.ts
+++ b/src/conninfo.ts
@@ -1,5 +1,4 @@
 import type { GetConnInfo } from 'hono/conninfo'
-import type { AddressInfo } from 'net'
 import type { HttpBindings } from './types'
 
 /**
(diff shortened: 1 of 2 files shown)


shogo-kurokawa-nm (approved):
Thank you for fixing the issue! LGTM!

yusukebe (approved):
LGTM!

yusukebe:
@nakasyou @shogo-kurokawa-nm Thanks! I'll merge this and release a new version.

PR #221, opened 2025-01-26
fix: Avoid error if connection is aborted before internal request object is created

usualoma:
Fixes https://github.com/honojs/hono/issues/1695

At this time, since the user is not listening for abort events, so we should return immediately.

### DIFF
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -1,7 +1,7 @@
 import type { IncomingMessage, ServerResponse, OutgoingHttpHeaders } from 'node:http'
 import type { Http2ServerRequest, Http2ServerResponse } from 'node:http2'
 import {
-  getAbortController,
+  abortControllerKey,
   newRequest,
   Request as LightweightRequest,
   toRequestError,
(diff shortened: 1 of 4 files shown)


usualoma:
@yusukebe 
Would you please review?

yusukebe:
Hey @usualoma 

https://github.com/honojs/hono/issues/1695 was fixed by #223. But do we still need to merge this PR?

usualoma:
Hi @yusukebe 

I think it would be better to merge this PR.

It may not be a fatal error, but as in the test case, if the request becomes unstable very early after the connection, and ends with a timeout, the error still occurs.

With this PR change, the existing standard case has no effect (no breaking change), so I think it would be better to merge it to avoid the error.

https://github.com/honojs/node-server/pull/221/files#diff-0d6c5bc426d150635c815993be8fb62aa429b7f22f5351adc7efd6f656ab3598R278-R287

yusukebe (approved):

yusukebe:
@usualoma 

Thank you so much for the explanation! I understood. Merging and shipping a new version!

PR #222, opened 2025-01-27
fix: export ServerOptions type

aryasaatvik:
Makes the ServerOptions type definition publicly accessible by adding the export keyword, allowing it to be imported and used by other modules that depend on this package.


yusukebe:
@aryasaatvik 

Thanks!

PR #223, opened 2025-01-30
fix: re-read request body from `incoming.rawBody` if available

usualoma:
In some environments (e.g. firebase functions), the body is already consumed, but we can re-read the request body from `incoming.rawBody` in that case.

fix https://github.com/honojs/hono/issues/1695

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -74,8 +74,19 @@ const newRequestFromIncoming = (
   }
 
   if (!(method === 'GET' || method === 'HEAD')) {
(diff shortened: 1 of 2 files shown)


usualoma:
@yusukebe
Would you please review?

yusukebe (approved):
LGTM!

yusukebe:
Hi @usualoma 

This fix is interesting. And the #1695 is completely fixed by this! I'll merge it now. Thank you so much!

Issue #230, opened 2025-03-18
serve-static.ts imports from `fs` vs. `node:fs` causing bundler warnings and potential deployment issues


firxworx:
## Issue Summary

All of the modules in `honojs/node-server` use the `node:` specifier prefix _except_ ` serve-static.ts` which imports from `fs` vs. `node:fs`.

https://github.com/honojs/node-server/blob/main/src/serve-static.ts

```ts
import { createReadStream, lstatSync } from 'fs'
import type { ReadStream, Stats } from 'fs'
```

These should be imported from `node:fs`.

## Impact

The omission causes warnings with leading bundlers such as Vite:

```
[plugin vite:resolve] Automatically externalized node built-in module "fs" imported from "node_modules/.pnpm/@hono+node-server@1.13.8_hono@4.7.4/node_modules/@hono/node-server/dist/serve-static.mjs". Consider adding it to environments.ssr.external if it is intended.
```

This can impact use on hosting environments such as Cloudflare which _"requires the use of the `node:` specifier prefix"_ when deploying workers that use node features with the `nodejs_compat` compatibility flag. Source: https://blog.cloudflare.com/workers-node-js-asynclocalstorage/

An example codebase hit by this issue is react-router v7 (remix) with hono replacing the default express backend. 

## Suggestion

Lint rules can be adjusted to enforce use of the `node:` specifier to prevent this sort of inconsistency/issue in future.

PR #231, opened 2025-03-18
fix: #230 use node:fs specifier prefix vs. fs import

firxworx:
Fix for https://github.com/honojs/node-server/issues/230

Ran `bun run test -- ./test/serve-static.test.ts` and all tests pass.


yusukebe (approved):
LGTM!

yusukebe:
@firxworx 

Thanks! Merging.

PR #232, opened 2025-03-18
WIP: feat: accept absolute-form URI for request-target

### LINKED ISSUE #228: Bug: 400 response when the http request target is in absolute-form
RFC 9112 states that the request target may be in absolute-form. (ref. https://datatracker.ietf.org/doc/html/rfc9112#name-request-target) This is used in particular when creating a proxy server.

When `@hono/node-server` and `hono` receives a request with an http request containing a request target in absolute-form, it will fail to handle the request and unintentionally cause a 400 error.

## Minimal reproducible example

```javascript
import { Hono } from "hono";

const app = new Hono();
app.all("*", () => new Response("ok"));
serve({
  fetch: app.fetch,
  port: 8080,
});
```

Then execute the following:

```bash
curl http://localhost:8081/foo -x http://localhost:8080 --verbose
```

This command will yield: (note that http://localhost:8081/foo is arbitrary and can be any url)

```
* Host localhost:8080 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:8080...
* Connected to localhost (::1) port 8080
> GET http://localhost:8081/foo HTTP/1.1
> Host: localhost:8081
> User-Agent: curl/8.6.0
> Accept: */*
> Proxy-Connection: Keep-Alive
> 
< HTTP/1.1 400 Bad Request
< content-type: text/plain; charset=UTF-8
< Date: Thu, 13 Mar 2025 10:25:29 GMT
< Connection: keep-alive
< Keep-Alive: timeout=5
< Transfer-Encoding: chunked
< 
* Leftovers after chunking: 5 bytes
* Connection #0 to host localhost left intact
```

usualoma:
fixes #228

* [ ]  Add tests
* [ ]  Run tests


usualoma:
@yusukebe
Would you please review this?

yusukebe (commented):
LGTM!

yusukebe:
Hi @usualoma 

I think this is perfect. Let's go with this. Thanks!

PR #235, opened 2025-04-05
fix: loading static files

houssems:
Hi,

I'm working in Angular 19 SSR integration with Hono. While investigating a server crash on the loading of static files, I discovered that it didn't wait for the res object to get resolved. The issue is from instanceof Promise.

`RangeError [ERR_HTTP_INVALID_STATUS_CODE]: Invalid status code: undefined at ServerResponse.writeHead (node:_http_server:351:11) at Hn (/dist/server/server.mjs:89:11) at Server.<anonymous> (/dist/server/server.mjs:91:554) at Server.emit (node:events:519:28) at parserOnIncoming (node:_http_server:1141:12) at HTTPParser.parserOnHeadersComplete (node:_http_common:118:17) { code: 'ERR_HTTP_INVALID_STATUS_CODE' }`

### DIFF
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -67,7 +67,7 @@ const responseViaResponseObject = async (
   outgoing: ServerResponse | Http2ServerResponse,
   options: { errorHandler?: CustomErrorHandler } = {}
 ) => {
-  if (res instanceof Promise) {
+  if (res != null && typeof res.then === 'function') {
     if (options.errorHandler) {
       try {
         res = await res


usualoma:
Hi @houssems. Thank you for creating a pull request!


### Should we accept thenable?

When the following code was executed in Deno and "wrangler," a message was displayed. Bun caused an error.

```ts
import { Hono } from 'hono'

const app = new Hono()

app.get('/', () => {
  const res = {
    then(resolve) {
      setTimeout(() => {
        resolve(new Response('Hello from thenable!'))
      }, 1000);
    }
  }
  return res as Promise<Response>
})

export default app
```

As the type indicates, Hono is originally expecting an instance of "Promise", but I think it's okay to accept a thenable even if it's not an instance of "Promise".


### `res != null`

Since it has already been guaranteed that res is not null or undefined using the following code before this point, I would like to omit `res != null` and prioritize performance.

https://github.com/honojs/node-server/blob/99ad1703fd3aa60853fbdfab880271ac88343016/src/listener.ts#L205

PR #236, opened 2025-04-10
fix: Handle Response Errors Correctly

jpulec:
It looks like during a refactor this logic got changed incorrectly. You need to await the promise in this catch block. Otherwise any error thrown inside `responseViaResponseObject` won't get caught and will bubble out the top.

### DIFF
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -224,7 +224,7 @@ export const getRequestListener = (
     }
 
     try {
-      return responseViaResponseObject(res, outgoing, options)
+      return await responseViaResponseObject(res, outgoing, options)
     } catch (e) {
       return handleResponseError(e, outgoing)
     }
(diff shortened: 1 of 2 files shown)


usualoma:
Hi @jpulec. Thank you for creating a pull request.

I see. There should be an await to supplement asynchronous processing errors. It would be great if you could add the test code. Is that possible?

jpulec:
There's a minimal test, based on the issue I was experiencing. Best I can tell, if your fetchCallback returns an object that might be kinda like a response, i.e. has a body property, but doesn't implement `.arrayBuffer()` for example. it would cause an unhandled promise to be thrown. This patch should resolve that.

usualoma:
Hi @jpulec. Thank you. I think it's good!

Please execute the following command.

```bash
$ npm run format:fix && npm run lint:fix
```

yusukebe (approved):
LGTM!

PR #241, opened 2025-05-11
feat: enable to get internal "getResponseState" function at Node.js v24.x

### LINKED ISSUE #240: Failed to find Response internal state key
### What version of Hono are you using?
- hono@4.7.8
- @hono/node-server@1.14.1

### What runtime/platform is your app running on? (with version if possible)
Node.js v24.0.0 on macOS

### What steps can reproduce the bug?
- Use Node.js v24.0.0
- Import @hono/node-server

```js
```

### What is the expected behavior?
No warning is printed.

### What do you see instead?
`Failed to find Response internal state key`

### Additional information
Commit in Undici: https://github.com/nodejs/undici/commit/45dceeaf376325df0fee3de83bc3d9f1e82a2cfd

usualoma:
fixes #240

With this PR, you will be able to apply optimisations to access internal data in v24 by executing the following.

```bash
$ node --import @hono/node-server/setup ./app.mjs
```

### What is this?

As pointed out by @timokoessler in #240, in Node.js v24.x, due to changes in https://github.com/nodejs/undici/commit/45dceeaf376325df0fee3de83bc3d9f1e82a2cfd, it is no longer possible to retrieve internal data using the previous node-server method.

Even when internal data cannot be retrieved, node-server continues to function normally; however, this may result in performance degradation for certain applications.

### Another way to get internal data

This change to ‘undici’ has made it more difficult to obtain than before, but there are ways to do so. If you execute the following code snippet in v24, you will see that it can be obtained.

```js
let getResponseState = null
const originalDeleteProperty = Reflect.deleteProperty
Reflect.deleteProperty = (target, prop) => {
  if (prop === 'getResponseState') {
    getResponseState = target.getResponseState
    Reflect.deleteProperty = originalDeleteProperty
  }
  return originalDeleteProperty(target, prop)
}

const res = new Response('Response Text')
console.log(getResponseState(res).body.source)
```

### The difficulty of making this code work

In Node.js, the Response class is dynamically initialised when `global.Response` is referenced.
This means that the code snippet above must be executed before `global.Response` is referenced for the first time. However, when executing `.mts` or `.mjs` files, `global.Response` is initialised during import resolution in most applications, so simply writing the code inside node-server will not work as expected.
To ensure that it works correctly, users must explicitly specify the `--import` option at runtime.

```bash
$ node --import @hono/node-server/setup ./app.mjs
```

### The option of giving up this optimisation

This optimisation was implemented in #145, but even without it, the speed in the ‘best case’ does not decrease. Therefore, I think there is an option to avoid making it as complex as in this PR and instead stop the optimisation that directly retrieves internal data to speed things up.

(However, if there is code that modifies the response headers, it will not achieve the ‘best case’ scenario and will remain a target for this optimisation, so I believe this optimisation applies to a fairly large number of applications.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,11 @@
       "types": "./dist/conninfo.d.ts",
       "require": "./dist/conninfo.js",
       "import": "./dist/conninfo.mjs"
+    },
+    "./setup": {
+      "types": "./dist/setup.d.ts",
+      "require": "./dist/setup.js",
+      "import": "./dist/setup.mjs"
     }
   },
   "typesVersions": {
--- a/src/response.ts
+++ b/src/response.ts
@@ -19,7 +19,7 @@ export class Response {
   #body?: BodyInit | null
   #init?: ResponseInit;
 
-  [getResponseCache](): typeof GlobalResponse {
+  [getResponseCache](): globalThis.Response {
     delete (this as any)[cacheKey]
     return ((this as any)[responseCache] ||= new GlobalResponse(this.#body, this.#init))
   }
@@ -89,7 +89,7 @@ if (!stateKey) {
 }
 
 export function getInternalBody(
-  response: Response | typeof GlobalResponse
+  response: Response | globalThis.Response
 ): InternalBody | undefined {
   if (!stateKey) {
     return
--- a/src/response.ts
+++ b/src/response.ts
@@ -3,12 +3,8 @@
 
 import type { OutgoingHttpHeaders } from 'node:http'
 import { buildOutgoingHttpHeaders } from './utils'
-
-interface InternalBody {
-  source: string | Uint8Array | FormData | Blob | null
-  stream: ReadableStream
-  length: number | null
-}
+import { getResponseState } from './utils/internal'
+import type { InternalBody } from './utils/internal'
 
 const responseCache = Symbol('responseCache')
 const getResponseCache = Symbol('getResponseCache')
@@ -81,25 +77,12 @@ export class Response {
 Object.setPrototypeOf(Response, GlobalResponse)
 Object.setPrototypeOf(Response.prototype, GlobalResponse.prototype)
 
-const stateKey = Reflect.ownKeys(new GlobalResponse()).find(
-  (k) => typeof k === 'symbol' && k.toString() === 'Symbol(state)'
-) as symbol | undefined
-if (!stateKey) {
-  console.warn('Failed to find Response internal state key')
-}
-
 export function getInternalBody(
   response: Response | globalThis.Response
 ): InternalBody | undefined {
-  if (!stateKey) {
-    return
-  }
-
   if (response instanceof Response) {
     response = (response as any)[getResponseCache]()
   }
 
-  const state = (response as any)[stateKey] as { body?: InternalBody } | undefined
-
-  return (state && state.body) || undefined
+  return getResponseState(response as globalThis.Response)?.body
 }
--- a/src/setup.ts
+++ b/src/setup.ts
@@ -1,0 +1,14 @@
+import { setGetResponseStateFn } from './utils/internal'
+
+const originalDeleteProperty = Reflect.deleteProperty
+export let getResponseState: Parameters<typeof setGetResponseStateFn>[0] | undefined
+Reflect.deleteProperty = (target, prop) => {
+  if (prop === 'getResponseState') {
+    getResponseState = (target as { getResponseState: Parameters<typeof setGetResponseStateFn>[0] })
+      .getResponseState
+    setGetResponseStateFn(getResponseState)
+    Reflect.deleteProperty = originalDeleteProperty
+  }
+  return originalDeleteProperty(target, prop)
+}
+Response
--- a/src/utils/internal.ts
+++ b/src/utils/internal.ts
@@ -1,0 +1,39 @@
+export interface InternalBody {
+  source: string | Uint8Array | FormData | Blob | null
+  stream: ReadableStream
+  length: number | null
+}
+
+// XXX: share "getResponseState" function by symbols via the global object to ensure more reliable.
+let getResponseStateFn: (res: Response) => { body: InternalBody } | undefined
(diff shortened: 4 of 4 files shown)


usualoma:
Hi @yusukebe,

I believe that the only way to resolve this issue is to proceed as follows.

```bash
$ node --import @hono/node-server/setup ./app.mjs
```

Do you think it is necessary to optimise in this complex manner?

The `--import` option is not necessary for those who do not require optimisation, so it may be best to leave it as it is, to be used only by those who wish to enable it.

yusukebe:
@usualoma 

Sorry for being late. I'm a little busy now and will comment on it later. Thanks.

yusukebe:
Hi @usualoma !

I am taking several benchmarks. I want to confirm in which cases there is a difference between using `--import @hono/node-server/setup` and not using it. Is there a difference in the Hello World app?

usualoma:
Hi @yusukebe, Thank you for your comment.

We can see the difference in the following application. (The key point is `res.headers.set('x-foo', 'bar')`.)

```ts
import { Hono } from 'hono'

const app = new Hono()

app.get('/', () => {
    const res = new Response('foo')
    res.headers.set('x-foo', 'bar')
    return res
})

serve({
    fetch: app.fetch,
    port: 3003,
})
```

### with `--import @hono/node-server/setup`

```
$ docker run -v `pwd`:/app -w /app/hono -p 3003:3003 --rm -it node:24 node --import @hono/node-server/setup ./app.mjs
```

```
$ bombardier -d 10s --fasthttp http://localhost:3003/
Bombarding http://localhost:3003/ for 10s using 125 connection(s)
[====================================================================================] 10s
Done!
Statistics        Avg      Stdev        Max
  Reqs/sec     48408.15    5503.67   57826.03
  Latency        2.58ms     1.42ms   138.00ms
  HTTP codes:
    1xx - 0, 2xx - 483930, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    11.03MB/s
```

### without `--import @hono/node-server/setup`

```
$ docker run -v `pwd`:/app -w /app/hono -p 3003:3003 --rm -it node:24 node ./app.mjs
```

```
$ bombardier -d 10s --fasthttp http://localhost:3003/
Bombarding http://localhost:3003/ for 10s using 125 connection(s)
[====================================================================================] 10s
Done!
Statistics        Avg      Stdev        Max
  Reqs/sec     42615.06    6739.75   52799.27
  Latency        2.93ms     3.17ms   332.86ms
  HTTP codes:
    1xx - 0, 2xx - 426148, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:     9.71MB/s
```

yusukebe:
Hi @usualoma 

Thank you for the answer! This is a little bit difficult to decide. But it's okay to merge this PR!

I also took a benchmark. I can't say it too loudly, but... I believe the most important aspect of benchmarks is ***comparing them with other frameworks***. In practical use, the difference may not be significant, but when compared with other frameworks, it becomes important.

In the current `hono` implementation, creating the instance of `Response` and setting the headers after it is not a common case. We can use `c.header()` or `c.text()` to avoid this. In the benchmark case, we can use `c.header()` or `c.text()`.

Also, if users have problems in the real world, I think it's a good idea to use `--import`.

Can you make this PR ready and ping me?

---

The following are just references. I created an app that performs header addition using Hono, Express, and Fastify, and took benchmarks.

```ts
// Hono
app.get('/', (c) => {
  const res = new Response('hi')
  res.headers.set('x-foo', 'bar')
  return res
})
```

```js
// Express
import express from 'express'

const app = express()
const port = 3001

app.get('/', (req, res) => {
  res.set('x-foo', 'bar')
  res.send('H!')
})

app.listen(port)
```

```js
// Fastify
import Fastify from 'fastify'

const fastify = Fastify({ logger: false })

fastify.get('/', async (request, reply) => {
  reply.header('x-foo', 'bar')
  return 'H!'
})

fastify.listen({ port: 3002 })
```

The result is as follows. If you don't include `--import`, it will lose to Express.

CleanShot 2025-05-17 at 05 20 54@2x

The following is an example using `c.text()`.

```ts
app.get('/', (c) => {
  c.header('x-foo', 'bar')
  return c.text('H!')
})
```

CleanShot 2025-05-17 at 05 34 34@2x

As expected, `c.text()` is fast.

usualoma:
Hi @yusukebe, Thank you for reviewing and confirming the performance!


### Is it common for headers to be modified?

As @yusukebe  mentioned, returning `c.text('H!')` in the handler avoids changing the header in many cases. However, when using middleware (from the perspective of Hono and the application, not node-server), it is often reasonable to rewrite the header. Even  the middleware included in Hono, approaches that rewrite headers such as `secure-headers` and `etag` are adopted.


### Headers and status codes are frequently referenced.

Middleware probably mainly references (and modifies) `headers` and `status`. For these two properties, it seems possible to prevent performance degradation (creation of `globalThis.Response` objects) when referenced, as shown in the following PR.

https://github.com/usualoma/node-server/pull/3


### It may not be necessary to refer to the internal data structure.

Even if we refer to `headers` and `status`, if we can still achieve a fast response, I feel that hacks such as `getInternalBody` are unnecessary.

https://github.com/honojs/node-server/pull/242


### Current thoughts

I have been considering this PR to apply the same approach as `getInternalBody` to v24, but in order to maintain stable behavior in future versions and achieve performance that does not require the use of `--import`, I believe it would be best to adopt https://github.com/honojs/node-server/pull/242.


@yusukebe I'm sorry to present another proposal. What do you think?

yusukebe:
Hi @usualoma !

#242 is great! I feel that the implementation of #242 seems to be simpler than the current implementation. I thought it could also cache the `url` and `statusText` properties, but they will not be referred to often, and the value will change depending on the implementation of `Request`. Shall we proceed with #242?

usualoma:
Hi @yusukebe 
Thanks for checking! Let's close this PR and move on to #242.

PR #243, opened 2025-05-22
fix: set `RequestError` name properly

### LINKED ISSUE #169: Issue with RequestError having static name property
Hard to say if this is a rollup issue or if this is a valid concern but the fix would be easy to implement.

When transpiling into esm. There is an issue with name being defined as a static value on the RequestError instead of just having it defined in the in the constructor. It throws an error saying that read only property name cannot be reassigned. 

Happy to open a PR if this would be approved since it wouldn't impact any existing functionality and it does seem odd to define a static field on a class that is defined on the prototype. The fix would be to move the name assignment to the constructor after the super is called.

yusukebe:
Fixes #169

We can't set the static property with `static name = 'RequestError'` in the `RequestError` class. The `name` property `RequestError` is already set.

CleanShot 2025-05-23 at 08 02 15@2x

So, I removed it and added a setter for the instance property.

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -7,7 +7,6 @@ import { Readable } from 'node:stream'
 import type { TLSSocket } from 'node:tls'
 
 export class RequestError extends Error {
-  static name = 'RequestError'
   constructor(
     message: string,
     options?: {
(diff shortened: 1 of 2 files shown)


yusukebe:
I'll merge this now. If you have any problems, let me know.

PR #245, opened 2025-06-06
feat: flush headers before sending data via readable stream

usualoma:
fixes: #244

Even if data is not immediately written from a ReadableStream, the header is sent without waiting for the data.

### Is it sufficient to only handle ServerResponse?

Yes, in the case of Http2ServerResponse, there is no need to explicitly flush because the header is sent without waiting for the data.

### Is this related to change #242?

No, it is not related. node-server has always behaved this way.

### DIFF
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -45,6 +45,14 @@ const handleResponseError = (e: unknown, outgoing: ServerResponse | Http2ServerR
   }
 }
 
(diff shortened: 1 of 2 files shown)


usualoma:
Hi @yusukebe!
Would you please review this?

yusukebe:
Hey @usualoma !

Great! I've confirmed that it works correctly. Thank you!

PR #249, opened 2025-07-03
fix: return promise from stream.cancel()

### LINKED ISSUE #248: Canceling an aborted stream terminates the hono process
Hi there,

we have an issue with our application running on node.js using hono.
The error we see in our logs is this one:

```
AbortError: The operation was aborted
    at destroyer (node:internal/streams/destroy:328:11)
    at Object.cancel (node:internal/webstreams/adapters:500:7)
    at invokePromiseCallback (node:internal/webstreams/util:172:10)
    at Object.cancelAlgorithm (node:internal/webstreams/util:177:23)
    at readableStreamDefaultControllerCancelSteps (node:internal/webstreams/readablestream:2414:37)
    at [kCancel] (node:internal/webstreams/readablestream:1089:12)
    at readableStreamCancel (node:internal/webstreams/readablestream:2018:39)
    at ReadableStream.cancel (node:internal/webstreams/readablestream:324:12)
    at writeFromReadableStream (./node_modules/.pnpm/@hono+node-server@1.14.1_hono@4.7.7/node_modules/@hono/node-server/dist/index.mjs:170:12)
    at responseViaResponseObject (./node_modules/.pnpm/@hono+node-server@1.14.1_hono@4.7.7/node_modules/@hono/node-server/dist/index.mjs:393:15)
```

After debugging this, I've noticed that the `cancel()` method of `ReadableStream` returns a promise.
So what we see here in the stack is a unhandled promise exception.

I will create a bugfix PR for this.

KnisterPeter:
The promise returned from stream.cancel() was ignored so far. This lead to unhandled promises when the stream throws during the cancel operation. In turn the whole hono node-server was shutdown.

Close #248

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -5,8 +5,7 @@ export function writeFromReadableStream(stream: ReadableStream<Uint8Array>, writ
   if (stream.locked) {
     throw new TypeError('ReadableStream is locked.')
   } else if (writable.destroyed) {
-    stream.cancel()
-    return
+    return stream.cancel()
   }
   const reader = stream.getReader()
   writable.on('close', cancel)
--- a/test/utils.test.ts
+++ b/test/utils.test.ts
@@ -1,4 +1,6 @@
-import { buildOutgoingHttpHeaders } from '../src/utils'
+import { console } from 'node:inspector/promises'
+import { Writable } from 'node:stream'
+import { buildOutgoingHttpHeaders, writeFromReadableStream } from '../src/utils'
 
 describe('buildOutgoingHttpHeaders', () => {
   it('original content-type is preserved', () => {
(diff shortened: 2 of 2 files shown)


yusukebe:
Hi @KnisterPeter 

Thank you for the PR. Can you add a test to confirm this change fixes the error?

KnisterPeter:
I will try my best 😃

KnisterPeter:
@yusukebe finally (after vacation) I managed to add a test case. Can you review it?

KnisterPeter:
@yusukebe I think the errors are not related to my changes. Can you confirm this?

yusukebe:
Hi @usualoma 

Do you think this change is needed?

usualoma:
Hi @yusukebe 
With the change in #258,  `cancel()` is no longer called, so if the problem does not recur in @KnisterPeter's environment on v1.17.0, I think this PR is unnecessary.

If there are other issues, I think we need to consider another approach.

Hi @KnisterPeter 
Could you please check how your environment has changed in v1.17.0?

KnisterPeter:
I will check and report back here.
Thanks for the explanation.

KnisterPeter:
@yusukebe @usualoma From our point of view the changes in #258 did fix our issue. Nice work 🥳 Thanks your for fixing it.
I'll close this PR and the issue by me.

yusukebe:
@KnisterPeter 

Woooow. Great! Thank you!!

PR #250, opened 2025-07-06
[WIP] feat: handle unconsumed or cancelled request bodies

### LINKED ISSUE #246: Client (web browser) hangs if body not consumed
When trying to abort POST request when using Hono with node-server, it is impossible to abort without consuming the whole body (which is wasteful, especially for large uploads).

Tried with node versions:
v22.17.0
v24.3.0 

Hono 4.8.2
Hono 4.8.3

Below is the simplest example:
```typescript
import { HTTPException } from "hono/http-exception";
import { Hono } from "hono";

const app = new Hono();

// Works
app.post(
  "/upload1",
  async (ctx) => {
    if (!ctx.req.raw.body) {
      throw new HTTPException(400, { message: "No file data received" });
    }
    // Create "/dev/null" writeable stream, without this the client will hang
    const nullStream = createWriteStream("/dev/null");
    ctx.req.raw.body.pipeTo(Writable.toWeb(nullStream))

    return ctx.json({ ok: "ok" });
  },
);

// Works
app.post(
  "/upload2",
  async (ctx) => {
    return ctx.json({ ok: "ok" });
  },
);

// Hangs
app.post(
  "/upload3",
  async (ctx) => {
    if (!ctx.req.raw.body) {
      throw new HTTPException(400, { message: "No file data received" });
    }

    return ctx.json({ ok: "ok" });
  },
);

// Hangs
app.post(
  "/upload4",
  async (ctx) => {
    if (!ctx.req.raw.body) {
      throw new HTTPException(400, { message: "No file data received" });
    }
    ctx.req.raw.body.cancel();
    return ctx.json({ ok: "ok" });
  },
);

serve(
  {
    fetch: app.fetch,
    port: 3000,
  },
  (info) => {
    console.log(`Server is running on http://localhost:${info.port}`);
  },
);
```

I suspect it is because of the how it is done in the following part, but I am speculating here.

https://github.com/honojs/node-server/blob/0e35e65480f972d15982ccaa2b8c813c7cdf7da6/src/request.ts#L88

usualoma:
fixes #246

### DIFF
--- a/test/server.test.ts
+++ b/test/server.test.ts
@@ -28,6 +28,40 @@ describe('Basic', () => {
   app.post('/posts', (c) => {
     return c.redirect('/posts')
   })
+  app.post('/no-body-consumed', (c) => {
+    if (!c.req.raw.body) {
+      // force create new request object
+      throw new Error('No body consumed')
+    }
+    return c.text('No body consumed')
+  })
+  app.post('/body-cancelled', (c) => {
+    if (!c.req.raw.body) {
+      // force create new request object
+      throw new Error('No body consumed')
+    }
+    c.req.raw.body.cancel()
+    return c.text('No body consumed')
+  })
+  app.post('/partially-consumed', async (c) => {
+    if (!c.req.raw.body) {
+      // force create new request object
+      throw new Error('No body consumed')
+    }
+    const reader = c.req.raw.body.getReader()
+    await reader.read() // read only one chunk
+    return c.text('No body consumed')
+  })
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @usualoma 

Is this ready for review?

usualoma:
Hi @yusukebe

Sorry. Actually, after thinking about it, I realized that this approach is not correct, so I am preparing a different version. Please wait a moment.

yusukebe:
@usualoma Okay!

PR #252, opened 2025-07-13
feat: Clean up the incoming object if the request is not completely finished.

usualoma:
fixes https://github.com/honojs/node-server/issues/246


usualoma:
Hi @yusukebe 
Would you please review this?

src/request.ts
```diff
@@ -83,6 +85,23 @@ const newRequestFromIncoming = (
           controller.close()
         },
       })
+    } else if ((incoming as IncomingMessageWithWrapBodyStream)[wrapBodyStream]) {
+      let reader: ReadableStreamDefaultReader<Uint8Array> | undefined
```
yusukebe, line 89: The current `ReadableStreamDefaultReader` is from `lib.dom`. Therefore, it throws type errors due to the type mismatch.

CleanShot 2025-07-14 at 05 44 43@2x

I think it's a good idea to use `ReadableStreamDefaultReader` imported from `node:stream/web`.

```diff
diff --git a/src/request.ts b/src/request.ts
index 31f378f..61664fb 100644
--- a/src/request.ts
+++ b/src/request.ts
@@ -4,6 +4,7 @@
 import type { IncomingMessage } from 'node:http'
 import { Http2ServerRequest } from 'node:http2'
 import { Readable } from 'node:stream'
+import type { ReadableStreamDefaultReader } from 'node:stream/web'
 import type { TLSSocket } from 'node:tls'

 export class RequestError extends Error {
```

yusukebe:
Hi @usualoma ! Thank you for the hard work!

I left a comment. Other than that, I think it looks good.

> Further investigation may reveal a solution, but since v18 has already reached end-of-life (LTS), I do not believe it is necessary to pursue this further. (For `node-server` as a whole, it may be advisable to remove v18 in the next major version.)

Indeed, we can end support for Node.js v18 in the next major version. I've created issue #253

yusukebe:
@usualoma 

It looks good to me. Thank you very much!

PR #255, opened 2025-07-14
docs: separate description of autoCleanupIncoming

usualoma:
Sorry for the minor change, I wanted to split the paragraphs, but they weren't separated in the first PR.

### before

CleanShot 2025-07-14 at 12 38 51@2x

### after

CleanShot 2025-07-14 at 12 38 02@2x


yusukebe:
Ah, I see. Thanks!

PR #258, opened 2025-07-15
fix: handle client disconnection without canceling stream

### LINKED ISSUE #239: `Invalid state: Controller is already closed`
### What version of Hono are you using?

4.7.6

### What runtime/platform is your app running on? (with version if possible)

node v20.14.0

### What steps can reproduce the bug?

- Create an handler that sends a Response with a `ReadableStream` body
- Close it on client side before it finishes

Repro: https://github.com/magne4000/repro-hono-streaming
`pnpm dev` then open `/`, refresh in quick succession, this will cause the error.

### What is the expected behavior?

This should not lead to a crash

### What do you see instead?

If a client closes the connection, it should be handled quietly

### Additional information

```
TypeError [ERR_INVALID_STATE]: Invalid state: Controller is already closed
    at ReadableStreamDefaultController.enqueue (node:internal/webstreams/readablestream:1074:13)
    at Timeout.pushRandomData (/home/magne/workspace/repro/hono-streaming/src/index.ts:38:20)
    at listOnTimeout (node:internal/timers:573:17)
    at process.processTimers (node:internal/timers:514:7) {
  code: 'ERR_INVALID_STATE'
}
```

Related issue using Hono with Vike: https://github.com/vikejs/vike/issues/2334

yusukebe:
Currently, if the client disconnects, the response stream will be canceled immediately:

https://github.com/honojs/node-server/blob/b5663ebfd7d6693fca2755c4f10d973d96eb9368/src/utils.ts#L8

https://github.com/honojs/node-server/blob/b5663ebfd7d6693fca2755c4f10d973d96eb9368/src/utils.ts#L21

I'm wondering what the correct behavior is in this situation, but it causes #239. To fix it, it should handle client disconnection gracefully.

Referring to https://github.com/hi-ogawa/reproductions/tree/main/web-to-node-handler-body-cancel, the behavior of `@mjackson/node-fetch-server` is different. With this PR, `@hono/node-server` behaves the same as it.

Fixes #239

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -5,38 +5,60 @@ export function writeFromReadableStream(stream: ReadableStream<Uint8Array>, writ
   if (stream.locked) {
     throw new TypeError('ReadableStream is locked.')
   } else if (writable.destroyed) {
-    stream.cancel()
     return
   }
+
   const reader = stream.getReader()
-  writable.on('close', cancel)
-  writable.on('error', cancel)
-  reader.read().then(flow, cancel)
+  let clientDisconnected = false
+
+  const handleClientDisconnect = () => {
+    clientDisconnected = true
+  }
+
+  const handleError = () => {
+    clientDisconnected = true
+  }
+
+  writable.on('close', handleClientDisconnect)
+  writable.on('error', handleError)
+
+  reader.read().then(flow, handleStreamError)
+
   return reader.closed.finally(() => {
-    writable.off('close', cancel)
-    writable.off('error', cancel)
+    writable.off('close', handleClientDisconnect)
+    writable.off('error', handleError)
   })
+
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  function cancel(error?: any) {
-    reader.cancel(error).catch(() => {})
-    if (error) {
-      writable.destroy(error)
+  function handleStreamError(error: any) {
+    if (!clientDisconnected) {
+      if (error) {
+        writable.destroy(error)
+      }
     }
   }
+
   function onDrain() {
-    reader.read().then(flow, cancel)
+    if (!clientDisconnected) {
+      reader.read().then(flow, handleStreamError)
+    }
   }
+
   function flow({ done, value }: ReadableStreamReadResult<Uint8Array>): void | Promise<void> {
+    if (clientDisconnected) {
+      return
+    }
+
     try {
       if (done) {
         writable.end()
       } else if (!writable.write(value)) {
         writable.once('drain', onDrain)
       } else {
-        return reader.read().then(flow, cancel)
+        return reader.read().then(flow, handleStreamError)
       }
     } catch (e) {
-      cancel(e)
+      handleStreamError(e)
     }
   }
 }
--- a/test/utils.test.ts
+++ b/test/utils.test.ts
@@ -1,4 +1,5 @@
-import { buildOutgoingHttpHeaders } from '../src/utils'
+import { Writable } from 'node:stream'
+import { buildOutgoingHttpHeaders, writeFromReadableStream } from '../src/utils'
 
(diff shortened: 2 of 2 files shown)


yusukebe:
Hi @usualoma 

Sorry for pinging again. Can you review this?

usualoma:
Hi @yusukebe 

When `writable` is closed or an error occurs, `writable.write(value)` will return a falsy value. In other words, if we do not call the `stream.cancel()`, I think we can simplify it as follows.


```diff
diff --git i/src/utils.ts w/src/utils.ts
index c6839bc..a54a707 100644
--- i/src/utils.ts
+++ w/src/utils.ts
@@ -9,46 +9,30 @@ export function writeFromReadableStream(stream: ReadableStream<Uint8Array>, writ
   }
 
   const reader = stream.getReader()
-  let clientDisconnected = false
-
-  const handleClientDisconnect = () => {
-    clientDisconnected = true
-  }
-
   const handleError = () => {
-    clientDisconnected = true
+    // add an error handler to prevent abnormal termination, but simply ignore the error
   }
 
-  writable.on('close', handleClientDisconnect)
   writable.on('error', handleError)
 
   reader.read().then(flow, handleStreamError)
 
   return reader.closed.finally(() => {
-    writable.off('close', handleClientDisconnect)
     writable.off('error', handleError)
   })
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   function handleStreamError(error: any) {
-    if (!clientDisconnected) {
-      if (error) {
-        writable.destroy(error)
-      }
+    if (error) {
+      writable.destroy(error)
     }
   }
 
   function onDrain() {
-    if (!clientDisconnected) {
-      reader.read().then(flow, handleStreamError)
-    }
+    reader.read().then(flow, handleStreamError)
   }
 
   function flow({ done, value }: ReadableStreamReadResult<Uint8Array>): void | Promise<void> {
-    if (clientDisconnected) {
-      return
-    }
-
     try {
       if (done) {
         writable.end()
```

usualoma:
Alternatively, if you want to ensure that `reader.read()` is not called again when a problem occurs with writable, you could do the following.

```diff
diff --git i/src/utils.ts w/src/utils.ts
index c6839bc..c22d09d 100644
--- i/src/utils.ts
+++ w/src/utils.ts
@@ -9,43 +9,33 @@ export function writeFromReadableStream(stream: ReadableStream<Uint8Array>, writ
   }
 
   const reader = stream.getReader()
-  let clientDisconnected = false
-
-  const handleClientDisconnect = () => {
-    clientDisconnected = true
-  }
-
   const handleError = () => {
-    clientDisconnected = true
+    // add an error handler to prevent abnormal termination, but simply ignore the error
   }
 
-  writable.on('close', handleClientDisconnect)
   writable.on('error', handleError)
 
   reader.read().then(flow, handleStreamError)
 
   return reader.closed.finally(() => {
-    writable.off('close', handleClientDisconnect)
     writable.off('error', handleError)
   })
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   function handleStreamError(error: any) {
-    if (!clientDisconnected) {
-      if (error) {
-        writable.destroy(error)
-      }
+    if (error) {
+      writable.destroy(error)
     }
   }
 
   function onDrain() {
-    if (!clientDisconnected) {
+    if (!writable.writable) {
       reader.read().then(flow, handleStreamError)
     }
   }
 
   function flow({ done, value }: ReadableStreamReadResult<Uint8Array>): void | Promise<void> {
-    if (clientDisconnected) {
+    if (!writable.writable) {
       return
     }
```

yusukebe:
@usualoma 

Thank you for the comment! I think the former is okay without preventing calling `reader.read()` again. Can you review this PR again?

usualoma (approved):
Thank you. LGTM!

PR #261, opened 2025-07-19
improve serve-static function

usualoma:
### fix: enable to serve filename with double dots

As described in the hono core, filenames such as `foo..bar.txt` should be accepted.

https://github.com/honojs/hono/blob/530ab09ae10caf33903dfb677dff239df01d5ded/src/utils/filepath.test.ts#L13-L17

#### Security check

Since the only untrusted string is `c.req.path`, I think the check should only be done in the following location.

https://github.com/honojs/node-server/compare/main...usualoma:node-server:refactor-serve-static?expand=1#diff-85001ab5aae1b04893fe64f90842d5e368e62920e6655ccc1792aa4dff852794R73-R80

I don't think any invalid strings will be entered here.

https://github.com/honojs/node-server/compare/main...usualoma:node-server:refactor-serve-static?expand=1#diff-85001ab5aae1b04893fe64f90842d5e368e62920e6655ccc1792aa4dff852794L102-L106

#### Stop calling `resolve()`

If security checks have been completed, I don't think it's necessary to call `resolve()` in `serveStatic()`.

### DIFF
--- a/src/serve-static.ts
+++ b/src/serve-static.ts
@@ -72,7 +72,7 @@ export const serveStatic = <E extends Env = any>(
       // Prevent encoded path traversal attacks
       if (!optionPath) {
         const decodedPath = decodeURIComponent(rawPath)
-        if (decodedPath.includes('..')) {
+        if (/(?:^|[\/\\])\.\.(?:$|[\/\\])/.test(decodedPath)) {
           await options.onNotFound?.(rawPath, c)
           return next()
         }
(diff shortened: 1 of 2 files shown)


usualoma:
Hi @yusukebe
I would like to make this adjustment. What do you think?

yusukebe (approved):
LGTM!

yusukebe:
@usualoma 

> Since the only untrusted string is `c.req.path`

Makes sense. We must consider only the case of `c.req.path`.

The code is cleaner and shorter. Looks good! Thank you.

PR #265, opened 2025-07-29
fix(listener): In Node.js v24, some response bodies are not read to the end until the next task queue.

### LINKED ISSUE #264: `server.test.ts` fails on Node.js v24
After merging #262, the test `server.test.ts` fails on only Node.js v24: https://github.com/honojs/node-server/actions/runs/16546886371/job/46795876085

I'm not sure whether we need to fix the test or the code, but it should be addressed.

usualoma:
Fixes #264

I considered making the following additional commit, but since it would change the behavior a bit and I am concerned about the performance impact (and, in fact, the need to deal with more case-sensitive parts), I would like to complete the fix with only the first commit and not add anything in particular.

https://github.com/usualoma/node-server/compare/fix-wait-next-task-node24...usualoma:node-server:fix-wait-next-task-node24-2

### DIFF
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -139,6 +139,14 @@ const responseViaResponseObject = async (
         done = true
       })
       if (!chunk) {
+        if (i === 1 && resHeaderRecord['transfer-encoding'] !== 'chunked') {
+          // XXX: In Node.js v24, some response bodies are not read all the way through until the next task queue,
+          // so wait a moment and retry. (e.g. new Blob([new Uint8Array(contents)]) )
+          await new Promise((resolve) => setTimeout(resolve))
+          i--
+          continue
+        }
+
         // Error occurred or currentReadPromise is not yet resolved.
         // If an error occurs, immediately break the loop.
         // If currentReadPromise is not yet resolved, pass it to writeFromReadableStreamDefaultReader.


src/listener.ts
```diff
@@ -139,6 +139,14 @@ const responseViaResponseObject = async (
         done = true
       })
       if (!chunk) {
+        if (i === 1 && resHeaderRecord['transfer-encoding'] !== 'chunked') {
```
yusukebe, line 142: Is it better to add the other stream-styled headers like these?

* `x-accel-buffering: no`
* `content-type: text/event-stream`

usualoma: Thank you for your comment.

When `x-accel-buffering: no` or `content-type: text/event-stream` is specified, `transfer-encoding: chunked` is usually included, so I think `resHeaderRecord[‘transfer-encoding’] !== ‘chunked’` alone is sufficient.

https://github.com/honojs/hono/blob/530ab09ae10caf33903dfb677dff239df01d5ded/src/helper/streaming/sse.ts#L86-L87

Additionally, even if this condition check is not perfect, the performance degradation is minimal, so I believe it will not cause issues in the asynchronous response pattern.

usualoma:
Hi @yusukebe.

The content is still in the first version. Please review with this.

yusukebe:
@usualoma 

Looks good to me. I'll merge and release a new version. Thank you!

> but since it would change the behavior a bit and I am concerned about the performance impact (and, in fact, the need to deal with more case-sensitive parts), I would like to complete the fix with only the first commit and not add anything in particular.

Agree!

PR #267, opened 2025-08-02
fix(listener): Limit retries to a maximum of three.

### LINKED ISSUE #266: v1.18.0 breaks Server-Sent Events (SSE) real-time streaming - first chunk delayed by pre-read logic
### **Summary**
@hono/node-server v1.18.0 introduced a regression that breaks real-time Server-Sent Events (SSE) streaming. The new pre-read logic delays the first chunk delivery, causing browsers to show "no response" and breaking the real-time nature of SSE connections.

### **Affected Versions**
- **Working**: v1.17.1 and earlier
- **Broken**: v1.18.0 and later

### **Root Cause**
The v1.18.0 release introduced a change described as "feat: always respond `res.body`" which added pre-read logic that attempts to read the first 2 chunks of a ReadableStream before sending headers and data to the client.

**Problematic code in `src/listener.ts` (lines ~138-175):**
```typescript
if (res.body) {
  const reader = res.body.getReader()
  const values: Uint8Array[] = []
  
  // This pre-read logic breaks real-time SSE streaming
  for (let i = 0; i < 2; i++) {
    currentReadPromise = reader.read()
    const chunk = await readWithoutBlocking(currentReadPromise)
    // ... waits to read chunks before sending anything
  }
  
  // Headers/data only sent after pre-reading
  outgoing.writeHead(res.status, resHeaderRecord)
}
```

### **Impact**
1. **Browser DevTools shows "no response"**: SSE connections appear stuck/pending
2. **Delayed first events**: Users don't see SSE data until multiple chunks are buffered
3. **Breaks real-time applications**: Chat apps, live updates, progress indicators, etc.
4. **Production issue**: Affects any application using SSE with async data generation

### **Reproduction**
```typescript
// SSE endpoint that generates data asynchronously
app.get('/events', (c) => {
  const stream = new ReadableStream({
    async start(controller) {
      controller.enqueue('data: first event\n\n')
      
      // Any async operation (DB query, API call, etc.)
      await new Promise(resolve => setTimeout(resolve, 100))
      controller.enqueue('data: second event\n\n')
      controller.close()
    }
  })
  
  c.header('Content-Type', 'text/event-stream')
  return c.body(stream)
})
```

**Expected**: First event appears immediately in browser  
**Actual (v1.18.0)**: Browser shows "no response" until second event is ready

### **Suggested Fix**
Add a fast-path for SSE responses that bypasses the pre-read logic:

```typescript
// Fast-path for Server-Sent Events
const contentTypeHeader = resHeaderRecord["content-type"] as string | undefined
if (contentTypeHeader && /^text\/event-stream\b/i.test(contentTypeHeader) && res.body) {
  // Stream immediately without pre-reading
  outgoing.writeHead(res.status, resHeaderRecord)
  flushHeaders(outgoing)
  await writeFromReadableStream(res.body, outgoing)
  return
}
```

### **Environment**
- **Node.js**: Any version
- **Browser**: All browsers (Chrome DevTools most visible)
- **Use case**: Any SSE implementation with async data generation

This regression significantly impacts real-time web applications and should be prioritized for the next patch release.

usualoma:
fixes #266

If the application did not include `Transfer-Encoding: chunked`, depending on the timing of enqueue, it would loop indefinitely until the second chunk was read.

One way to do this is to treat `text/event-stream` as special, as @yusukebe pointed out in this comment, but I would prefer to resolve this without depending on a specific `Content-Type`.

https://github.com/honojs/node-server/pull/265#discussion_r2241271562

Essentially, the problem was simply [#265](https://github.com/honojs/node-server/pull/265)'s `i--` and `currentReadPromise = reader.read()`, so I think that fixing that should be sufficient without any additional logical changes.

### Minimal reproducible example

https://github.com/usualoma/hono-se-266

### DIFF
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -131,9 +131,10 @@ const responseViaResponseObject = async (
     let done = false
     let currentReadPromise: Promise<ReadableStreamReadResult<Uint8Array>> | undefined = undefined
 
-    // In the case of synchronous responses, usually a maximum of two readings is done
-    for (let i = 0; i < 2; i++) {
-      currentReadPromise = reader.read()
+    // In the case of synchronous responses, usually a maximum of two (or three in special cases) readings is done
+    let maxReadCount = 2
+    for (let i = 0; i < maxReadCount; i++) {
+      currentReadPromise ||= reader.read()
       const chunk = await readWithoutBlocking(currentReadPromise).catch((e) => {
         console.error(e)
         done = true
@@ -143,7 +144,7 @@ const responseViaResponseObject = async (
           // XXX: In Node.js v24, some response bodies are not read all the way through until the next task queue,
           // so wait a moment and retry. (e.g. new Blob([new Uint8Array(contents)]) )
           await new Promise((resolve) => setTimeout(resolve))
-          i--
+          maxReadCount = 3
           continue
         }
 
--- a/test/server.test.ts
+++ b/test/server.test.ts
@@ -137,6 +137,8 @@ describe('various response body types', () => {
     const largeText = 'a'.repeat(1024 * 1024 * 10)
     let server: ServerType
     let resolveReadableStreamPromise: () => void
+    let resolveEventStreamPromise: () => void
+    let resolveEventStreamWithoutTransferEncodingPromise: () => void
     beforeAll(() => {
       const app = new Hono()
       app.use('*', async (c, next) => {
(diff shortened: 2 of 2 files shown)


usualoma:
@yusukebe
Would you please review this?

yusukebe:
Hi @usualoma 

I've tried some edge cases based on your repro. They work well! Looks good, thank you!

PR #271, opened 2025-08-12
fix: Skip content-length assignment when transfer-encoding is chunked.

### LINKED ISSUE #270: Issue when using `@hono/mcp`, content-length is included when chunked encoding is used.
I was trying to use `@hono/mcp`, when using Inspector it returns a precarious error. 

```
[cause]: HTTPParserError: Response does not match the HTTP/1.1 protocol (Content-Length can't be present with Transfer-Encoding)
      at Parser.execute (node:internal/deps/undici/undici:5998:19)
      at Parser.readMore (node:internal/deps/undici/undici:5957:16)
      at Socket.<anonymous> (node:internal/deps/undici/undici:6285:18)
      at Socket.emit (node:events:518:28)
      at emitReadable_ (node:internal/streams/readable:834:12)
      at process.processTicksAndRejections (node:internal/process/task_queues:89:21) {
    code: 'HPE_UNEXPECTED_CONTENT_LENGTH',
    data: 'c9\r\n' +
      'event: message\n' +
      'data: {"result":{"protocolVersion":"2025-06-18","capabilities":{"tools":{"listChanged":true}},"serverInfo":{"name":"example-mcp-server","version":"1.0.0"}},"jsonrpc":"2.0","id":0}\n' +
      '\n' +
      '\r\n' +
      '0\r\n' +
      '\r\n'
  }
```

Upon further debugging I traced that down to this line, https://github.com/honojs/node-server/blob/main/src/listener.ts#L167-L169, to me it sees more reasonable to expand the logic to check for `chunked` perhaps expand the code to `shouldAddContentLength(headers): true/false` and then derive it.

Happy to open PR if ya'll think that's the best way :)

usualoma:
fixes #270

In `responseViaCache`, we avoided searching for `Transfer-Encoding` because it was computationally expensive when the header object passed was a `Record<string, string>`. However, in this case, we assume that the body is not chunked if it is a `string` or `Blob`, so we would like to process it based on that assumption. (This has actually been the behavior so far.)
In `responseViaResponseObject`, we had also avoided referencing `Transfer-Encoding` to align with `responseViaCache`, but since there are actual issues like #270, I would like to avoid setting `Content-Length` if `Transfer-Encoding` is chunked.

### DIFF
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -74,6 +74,7 @@ const responseViaCache = async (
     header = buildOutgoingHttpHeaders(header)
   }
 
+  // in `responseViaCache`, if body is not stream, Transfer-Encoding is considered not chunked
   if (typeof body === 'string') {
     header['Content-Length'] = Buffer.byteLength(body)
   } else if (body instanceof Uint8Array) {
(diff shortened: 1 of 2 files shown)


usualoma:
@yusukebe
Would you please review this?

yusukebe:
@usualoma 

Thank you for the quick fix. Looks good! Merging.

PR #273, opened 2025-08-17
feat: add log when directory does not exists

Its-Just-Nans:
Can be useful if you place a wrong `root` path :)


yusukebe:
Hi @Its-Just-Nans !

This is a good improvement! I've fixed the lint/format error by myself. Looks good. I'll merge and release a new version. Thanks!

Issue #269, opened 2025-08-04
Using hono-node with the cors middleware causes a memory leak


krzkz94:
### What version of Hono are you using?

4.8.12

### What runtime/platform is your app running on? (with version if possible)

Node

### What steps can reproduce the bug?

Clone this repo
https://github.com/krzkz94/hono-context-memory-leak

Follow the outlined steps
(install, run app, run autocannon)

Notice heap size increasing
```
npx tsx src/index.ts
Server is running on http://localhost:3000
{ before: '9.09', after: '9.09' }
{ before: '9.11', after: '9.11' }
{ before: '51.49', after: '51.50' }
{ before: '67.14', after: '67.14' }
{ before: '111.91', after: '111.91' }
{ before: '155.77', after: '155.77' }
{ before: '126.81', after: '126.81' }
{ before: '162.79', after: '162.79' }
{ before: '207.94', after: '207.94' }
{ before: '254.23', after: '254.23' }
{ before: '302.14', after: '302.14' }
{ before: '342.52', after: '342.52' }
{ before: '380.52', after: '380.52' }
{ before: '248.96', after: '248.96' }
{ before: '288.23', after: '288.23' }
{ before: '339.07', after: '339.07' }
{ before: '382.03', after: '382.03' }
{ before: '424.34', after: '424.34' }
{ before: '458.85', after: '458.85' }
{ before: '511.55', after: '511.55' }
{ before: '552.79', after: '552.79' }
{ before: '596.38', after: '596.38' }
{ before: '590.64', after: '590.64' }
```

Node22

### What is the expected behavior?

Heap size stays consistent with minimal increases, for example, below, it's the result _without_ the cors middleware.

```
tsx watch src/index.ts

Server is running on http://localhost:3000
{ before: '17.41', after: '17.41' }
{ before: '11.33', after: '11.33' }
{ before: '10.58', after: '10.58' }
{ before: '23.65', after: '23.65' }
{ before: '14.78', after: '14.78' }
{ before: '17.13', after: '17.13' }
{ before: '16.21', after: '16.21' }
{ before: '16.23', after: '16.23' }
{ before: '13.44', after: '13.45' }
{ before: '14.90', after: '14.90' }
{ before: '21.29', after: '21.29' }
{ before: '18.71', after: '18.71' }
{ before: '14.91', after: '14.91' }
{ before: '13.46', after: '13.46' }
```

### What do you see instead?

Heap size increases and stays high even after the test ends

```
[...]

PR #274, opened 2025-08-29
fix: Keep lightweight request even if accessing method, url or headers

usualoma:
#269

### What is this?

Stop generating heavy `global.Request` just by accessing `req.headers`.

This eliminates the memory consumption that until now was caused by simply applying the simple middleware called `cors()`, which generated a `global.Request` and was hard to free up until `GC` was executed.

### Cases not resolved

In situations where we accept a lot of lightweight POST requests, this pull request will not improve on memory usage.

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -42,14 +42,7 @@ export class Request extends GlobalRequest {
   }
 }
 
-export type IncomingMessageWithWrapBodyStream = IncomingMessage & { [wrapBodyStream]: boolean }
-export const wrapBodyStream = Symbol('wrapBodyStream')
-const newRequestFromIncoming = (
-  method: string,
-  url: string,
-  incoming: IncomingMessage | Http2ServerRequest,
-  abortController: AbortController
-): Request => {
+const newHeadersFromIncoming = (incoming: IncomingMessage | Http2ServerRequest) => {
   const headerRecord: [string, string][] = []
   const rawHeaders = incoming.rawHeaders
   for (let i = 0; i < rawHeaders.length; i += 2) {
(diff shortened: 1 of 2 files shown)


usualoma:
@yusukebe 

It won't solve every situation, but I believe it will improve things for very simple apps like the one below (which involves accessing headers).

https://github.com/yusukebe/hono-nodejs-adapter-memory-leak

Shall we add this?

src/request.ts
```diff
@@ -58,10 +51,21 @@ const newRequestFromIncoming = (
       headerRecord.push([key, value])
     }
   }
+  return new Headers(headerRecord)
```
yusukebe, line 54: Using `new Headers` instead of creating `Request` is nice.

It's just an idea, maybe overwork, we should not add it in this PR, but is creating `LightweightHeaders` a good idea?

```ts
class LightweightHeaders {
  constructor(rawHeaders) {} // or constructor(headerRecord)

  get(name) {
    // find the value from rawHeaders or headerRecord
  }

  has(name) {
    // find the value from rawHeaders or headerRecord
  }

  set(name, value) {
    this.toHeaders.set(name, value)
  }

  toHeaders() {
    const headers = new Headers(this.entries())
    // cache headers
    return headers
  }
}
```

It's for readonly, so it returns the value by `.get()` from `rawHeaders` or `headerRecord`, and if accessing `.set()`, do `new Headers()`.

Anyway, we can merge this PR.

yusukebe (approved):
LGTM!

yusukebe:
@usualoma 

> Shall we add this?

Yes! Thank you!

yusukebe:
I wondered if we should label `fix`, `feat`, or others, but I labeled it `fix`.

usualoma:
@yusukebe 

Thank you!

> I wondered if we should label fix, feat, or others, but I labeled it fix.

src/request.ts
usualoma: `LightweightHeaders` is a great idea! However, I'm concerned that using it like `fetch(url, {headers: c.req.raw.headers})` would introduce more compatibility issues to worry about, making it not worth the maintenance cost.

PR #277, opened 2025-09-12
fix: Refactor promise check for response handling

kay-is:
I got an error because `instanceof` wouldn't recognize `res` as a `Promise`. 

Might be related to my usage of `tsx`, idk if it breaks the prototype chain or something.

Anyway, if I change to check for a `then` method, everything works again.

### DIFF
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -103,7 +103,7 @@ const responseViaResponseObject = async (
   outgoing: ServerResponse | Http2ServerResponse,
   options: { errorHandler?: CustomErrorHandler } = {}
 ) => {
-  if (typeof res.then === 'function') {
+  if ("then" in res && typeof res.then === 'function') {
     if (options.errorHandler) {
       try {
         res = await res
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -103,7 +103,7 @@ const responseViaResponseObject = async (
   outgoing: ServerResponse | Http2ServerResponse,
   options: { errorHandler?: CustomErrorHandler } = {}
 ) => {
-  if (res instanceof Promise) {
+  if (typeof res.then === 'function') {
     if (options.errorHandler) {
       try {
         res = await res
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -98,12 +98,15 @@ const responseViaCache = async (
   ;(outgoing as OutgoingHasOutgoingEnded)[outgoingEnded]?.()
 }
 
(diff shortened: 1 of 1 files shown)


usualoma:
Oh, sorry, I accidentally approved it by mistake.

usualoma:
Hi @kay-is, thanks for creating the pull request.

I agree with the change to treat Thenable as a Promise.

I think the type error issue can be resolved like this.

```diff
diff --git i/src/listener.ts w/src/listener.ts
index 2724ffd..a9ea3dd 100644
--- i/src/listener.ts
+++ w/src/listener.ts
@@ -98,12 +98,15 @@ const responseViaCache = async (
   ;(outgoing as OutgoingHasOutgoingEnded)[outgoingEnded]?.()
 }
 
+const isPromise = (res: Response | Promise<Response>): res is Promise<Response> =>
+  typeof (res as Promise<Response>).then === 'function'
+
 const responseViaResponseObject = async (
   res: Response | Promise<Response>,
   outgoing: ServerResponse | Http2ServerResponse,
   options: { errorHandler?: CustomErrorHandler } = {}
 ) => {
-  if (res instanceof Promise) {
+  if (isPromise(res)) {
     if (options.errorHandler) {
       try {
         res = await res
```

Could you please add the tests? Then I think we can merge it.

kay-is:
@usualoma sure thing!

yusukebe:
Hi @kay-is, thank you for the PR.

> I got an error because `instanceof` wouldn't recognize `res` as a `Promise`. 

Can you reproduce it? If so, it's better to add a test to confirm that the issue will be resolved with this fix. If it's difficult, we can treat this change as a `refactor` with no tests.

kay-is:
@yusukebe sorry, no idea exactly *how* that error happened or how to reproduce it. 

I just had a codebase where `res` wasn't an instance of `Promise` anymore. However, it was still a promise, just didn't have its prototype associated with the `Promise` class. Even your static file server did produce such kind of promise objects, so I'm quite confused.

I assume it has something to do with the `tsx` package or my `tsconfig.json`. Like, it affects all code, even the third-party code, of my codebase. But I'm not 100% sure.

yusukebe (approved):
LGTM!

yusukebe:
@kay-is 

No problem! If so, let's go without tests. This is like a refactoring but resolving your issue. So, I'll release a new patch version after merging. Thank you for your contribution!

Issue #279, opened 2025-09-24
ReadableStream.cancel never trigger


unnoq:
`ReadableStream.cancel` should be triggered when the user cancels a response before it finishes.

Currently, only `"aborted"` is printted, while `"cancelled"` never prints when a response is canceled before completion.

`"@hono/node-server": "^1.19.4",`

```ts
import { serve } from "@hono/node-server";

serve({
    fetch: ({ signal }) => {
        signal?.addEventListener("abort", () => console.log("aborted"));

        return new Response(new ReadableStream({
            async pull(controller) {
                controller.enqueue(new TextEncoder().encode("Hello, World!"));
                await new Promise((resolve) => setTimeout(resolve, 1000));
            },
            async cancel() {
                console.log("cancelled");
            }
        }), {
            headers: {
                "content-type": "text/event-stream"
            }
        })
    },
    port: 3000,
});
```

usualoma:
Hi @unnoq, thank you for your comment. We will look into it.

We may need to review the details of the following changes.
https://github.com/honojs/node-server/pull/258

usualoma:
### What kind of usage is ReadableStream intended for?

As shown in the example below, it is expected that when an instance of ReadableStream no longer needs to generate data, cancel() is called, and upon being called, it should stop generating data itself.

https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/ReadableStream#examples


### Examples mentioned in #239

In the example mentioned below, an error occurs because the generation started within `start()` does not stop when `cancel()` is called.

https://github.com/honojs/node-server/issues/239

In response to this, [#258](https://github.com/honojs/node-server/pull/258) addressed the issue by preventing `cancel()` from being called. However, I believe the correct approach would have been to leave the code calling `cancel()` unchanged and instead modify the application side.


### How does it behave in other runtimes?

I have prepared the repository.

https://github.com/usualoma/hono-stream-cancel

#### app.ts

This is the implementation using `pull()` shown in the example for this issue.

When executed as follows, only node-server does not output "cancel".

```ts
# Confirmation of bun's behavior
$ bun run --hot app.ts &
$ curl -N http://localhost:3000/
```

```ts
# Confirmation of deno's behavior
$ deno serve app.ts &
$ curl -N http://localhost:8000/
```

```ts
# Confirmation of node-server's behavior
$ npx tsx app-node.ts &
$ curl -N http://localhost:3000/
```


#### app2.ts

[...]

PR #280, opened 2025-09-28
fix: cancel a readable stream if a writable stream is closed before a readable stream is closed.

usualoma:
fix: #279

There may be compatibility issues with other frameworks or existing implementations on different servers; however, calling `cancel()` appears to be the correct approach according to web standards.


yusukebe:
@usualoma 

I see this makes sense. Thank you for handling it.

PR #282, opened 2025-10-04
[WIP] feat: introduce `hono-node-server` script to serve Hono app

usualoma:
Starting with Node.js v24, experimental-strip-types is enabled by default. This allows us to run an app built with Hono in .ts format using very simple code and without external dependencies, simply by adding this PR.

While there are limitations, I find it convenient that you can run it in Node.js simply by changing the command, using exactly the same code as Deno or Bun.

```bash
$ cat app.ts
import { Hono } from 'hono'

const app = new Hono()
app.get('/', async (c) => {
  return c.json({ message: 'Hello World' })
})

export default app

$ npx @hono/node-server app.ts
Listening on http://localhost:3000
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,9 @@
   "description": "Node.js Adapter for Hono",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
+  "bin": {
+    "hono-node-server": "./dist/serve.js"
+  },
   "files": [
     "dist"
   ],
--- a/src/serve.ts
+++ b/src/serve.ts
@@ -1,0 +1,27 @@
+import { realpathSync } from 'node:fs'
+import { parseArgs } from 'node:util'
+import { serve } from './server'
+
+const { values, positionals } = parseArgs({
+  options: {
+    port: { type: 'string' },
+  },
+  allowPositionals: true,
+})
+
+if (positionals.length === 0) {
+  throw new Error('Please specify the path to the app file.')
+}
+
+const appFilePath = realpathSync(positionals[0])
+import(appFilePath).then(({ default: app }) => {
+  serve(
+    {
+      fetch: app.fetch,
+      port: values.port ? Number.parseInt(values.port) : undefined,
+    },
+    (info) => {
+      console.log(`Listening on http://localhost:${info.port}`)
+    }
+  )
+})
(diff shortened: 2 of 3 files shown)


usualoma:
Hi @yusukebe 

It's not exactly “super convenient,” but the ability to run on multiple runtimes with completely identical code feels Hono-like. What do you think?

yusukebe:
@usualoma 

Ah, interesting! Why have I never come up with this! I'll check the details later.

yusukebe:
Hey @usualoma !

I think we don't have to hurry to decide whether to accept this PR or not. We have to consider well. I'll share my thoughts about this feature:

### How about making this not depend on Hono

This script can be used for the app that uses the `fetch` interface, not Hono. Like this app:

```js
export default {
  fetch: () => new Response('Hi')
}
```

So, it may be good to design support for all `fetch` interface apps.

### I want to run JSX with this!

If we use this script, the app is usually run on Node.js. Now, it's wonderful that Node.js can run TypeScript! However, if the app uses JSX, it can't do so. I love the JSX feature of Hono. So it's better to work with JSX. I think we can do it with a builder like esbuild, though I don't know too much about the software for building. I have to investigate it.

But, to keep it simple, it will be better only to rely on Node.js type stripping.

### Is it best to create the script in this `@hono/node-server`?

There are several options for distributing this script, not just including it in this `@hono/node-server` package. If we make it so it does not depend on Hono, it will be more effective. For example, we can create `node-server-for-fetch-app` (this is done intentionally for clarity). If so, it can cover more general use-cases. And we can name the script. For example, how about `s` as super super short?

```
$ s src/index.ts
```

(I like the naming `serve`, but there is already the`serve` package: https://www.npmjs.com/package/serve)

### WinterTC started to consider the HTTP server API

This not only matters in Node.js matters, but, interestingly, WinterTC started to discuss the HTTP server API: https://github.com/WinterTC55/proposal-http-server-api/issues/1

If this is accepted, Node.js can follow the spec, and the interface will be implemented; therefore, we won't need to use this script. But it's a better idea to use this until it is enabled.

---

Anyway, this is an exciting idea.

yusukebe:
This is a similar idea: https://github.com/h3js/srvx

usualoma:
Hi @yusukebe, thanks for the comment.

Indeed, it would be great if JSX worked too.

It seems https://github.com/h3js/srvx uses https://github.com/unjs/jiti.

https://github.com/h3js/srvx/blob/main/src/cli.ts#L265

PR #283, opened 2025-10-19
feat: introduce new option requestsPerForcedGC

usualoma:
Normally, this setting is not required, but for environments where GC execution is expected to be delayed, an option has been added to explicitly trigger GC every N requests.

#### app

```ts
import { Hono } from 'hono'

const app = new Hono()

app.get('/', (c) => c.text('OK'))

serve(
  {
    fetch: app.fetch,
    requestsPerForcedGC: 1000, // Explicitly trigger garbage collection every 1000 requests.
  },
  (info) => {
    console.log(`Server is running on http://localhost:${info.port}`)
  }
)
```

#### runner

```bash
$ node --expose-gc dist/index.js
```

### DIFF
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -203,9 +203,14 @@ export const getRequestListener = (
     errorHandler?: CustomErrorHandler
     overrideGlobalObjects?: boolean
     autoCleanupIncoming?: boolean
+    requestsPerForcedGC?: number
   } = {}
 ) => {
   const autoCleanupIncoming = options.autoCleanupIncoming ?? true
+  const requestsPerForcedGC = options.requestsPerForcedGC ?? 0
+  let requestCount = 0
+  const gc = global.gc as () => void
+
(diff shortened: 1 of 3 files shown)


usualoma:
Hi @yusukebe

There's no guarantee that issue #269 will be resolved, but based on our investigation so far, I believe memory is being freed when `global.gc()` is explicitly called. How about we try implementing this change?

yusukebe:
I've been thinking about this for a long time. "Clearing something every certain number of requests" is making me think nostalgically of `MaxRequestsPerChild` in Apache. 

It may be a practical solution. But I wonder if this is a good way for "Node.js"? I checked, and Express doesn't have a function like that. So I am not sure this method is suitable for the Node.js Adapter. It's better to wait; we've got more reports about memory leaks.

usualoma:
Hi @yusukebe, thank you for the review.
Yes, I don't think it's an essential fix.

Ideally, we'd eliminate the memory increase caused by AbortSignal...

PR #286, opened 2025-10-30
fix(serve-static): fix onFound timing

### LINKED ISSUE #225: serveStatic: unable to override `content-type` header in `onFound`
/subj

## repro

1. `curl https://cataas.com/cat > static/meow`

2. run:
```
const app = new Hono()
app.use('/*', serveStatic({
  root: 'static',
  onFound: (path, ctx) => {
    ctx.res.headers.set('Content-Type', 'image/jpeg')
  }
})

serve(app)
```

3. `curl -I localhost:3000/meow`

## expected

`Content-Type` header is set to `image/jpeg`

## actual

`Content-Type` header is set to `application/octet-stream` because there is no extension

## additional context

https://github.com/honojs/node-server/blob/main/src/serve-static.ts#L108-L111 – onFound is called before the mime type is derived from the extension, so simply moving it a few lines down should solve this

usualoma:
Fix #225

As mentioned in #225, I also believe the timing at which `onFound` is called in `serveStatic` of `node-server` is somewhat premature.

In the hono core, it is called immediately before returning the response, so I think it should be aligned accordingly.

https://github.com/honojs/hono/blob/main/src/middleware/serve-static/index.ts#L117

### DIFF
--- a/src/serve-static.ts
+++ b/src/serve-static.ts
@@ -105,7 +105,6 @@ export const serveStatic = <E extends Env = any>(
       await options.onNotFound?.(path, c)
       return next()
     }
-    await options.onFound?.(path, c)
 
     const mimeType = getMimeType(path)
     c.header('Content-Type', mimeType || 'application/octet-stream')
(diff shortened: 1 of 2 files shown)


usualoma:
@yusukebe
Would you please review this?

yusukebe (approved):
LGTM!

yusukebe:
@usualoma 

Thank you! I've confirmed, and it works well. Merging.

PR #290, opened 2025-12-06
fix: Fix for hono issue 4563 - incorrect content-length after following symlink

tshmieldev:
This PR fixes https://github.com/honojs/hono/issues/4563
The issue was, lstat does not follow symlinks, so it read the wrong content length and an incomplete file was returned to the client. Changed lstat to stat, and the behavior is as expected (matches what we get when we run the server with bun adapters instead of node-server)

### DIFF
--- a/test/serve-static.test.ts
+++ b/test/serve-static.test.ts
@@ -198,7 +198,8 @@ describe('Serve Static Middleware', () => {
     expect(res.status).toBe(404)
   })
 
-  it('Should follow symlinks', async () => {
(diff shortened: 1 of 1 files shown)


usualoma:
Thank you 👍

tshmieldev:
Please let me know if this looks good or if you’d prefer some adjustments, I'm really happy to help! :)

yusukebe:
@tshmieldev 

Nice fix! Thank you very much for your contribution!

PR #293, opened 2026-01-07
fix(serve-static): Use Readable.toWeb in serveStatic

otya128:
The previous createStreamBody implementation used flowing mode. This caused a flaw where data would accumulate in the queue if the transmission speed was slower than the read speed, potentially consumig memory equal to the file size.

However, as noted in the commit 5064e925e78ab0dee1d5e8b10cc63f9b1c5929cf that introduced this issue, using Readable.toWeb may cause an exception in Node.js versions that do not have nodejs/node#54206 applied. Therefore, the old implementation remains. It is desirable that it will be removed when support for older versions of Node.js ends.

### DIFF
--- a/src/serve-static.ts
+++ b/src/serve-static.ts
@@ -3,6 +3,8 @@ import { getMimeType } from 'hono/utils/mime'
 import type { ReadStream, Stats } from 'node:fs'
 import { createReadStream, statSync, existsSync } from 'node:fs'
 import { join } from 'node:path'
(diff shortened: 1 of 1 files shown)


usualoma:
Hi @otya128, thank you. The changes look great!

yusukebe (approved):
LGTM!

yusukebe:
@otya128 Thanks! I'll merge this and release a new version.

PR #295, opened 2026-01-12
fix(globals): Stop overwriting global.fetch

### LINKED ISSUE #294: Library is overriding global.fetch on import
https://github.com/honojs/node-server/blob/main/src/globals.ts

This is super extra biggie bad. There is some discussion on #154 and some suggestion that it's fixed, but this globals.ts file doesn't care about any runtime parameters, it just clobbers global.fetch on load.

usualoma:
fix #294

`compress: false` was added below.

https://github.com/honojs/node-server/pull/41

This is likely for compatibility with the compress option when using node-fetch.

https://www.npmjs.com/package/node-fetch

The compress option is not used with the Fetch API.

https://developer.mozilla.org/docs/Web/API/RequestInit

Since we no longer need to consider the use of node-fetch in the current version of node-server, I believe we can stop overriding `global.fetch`.

### DIFF
--- a/src/globals.ts
+++ b/src/globals.ts
@@ -1,19 +1,6 @@
 import crypto from 'node:crypto'
 
-const webFetch = global.fetch
-
 /** jest dose not use crypto in the global, but this is OK for node 18 */
 if (typeof global.crypto === 'undefined') {
   global.crypto = crypto as Crypto
-}
-
-global.fetch = (info, init?) => {
-  init = {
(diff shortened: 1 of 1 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@usualoma 

I think it's better to mark it as a `fix` since we have to release a new patch version including this change. I've changed the title myself.

Thanks!

Issue #76, opened 2023-07-29
Performance improvement


yusukebe:
Indeed, the Hono app, when using node-server, is slow. You can see the benchmarks here:

https://github.com/SaltyAom/bun-http-framework-benchmark

However, there are opportunities to improve performance, and we should do them.

rafaell-lycan:
I agree, especially if it performs better than `Express` and `Hapi` or even as well as `Koa` or `Fastify`.

Nothing is perfect, but there's always room to improve 🔥

cdtut:
@yusukebe Is performance the only limitation of hono on node or are there missing functionality or feature compared to bun?

kevinvdburgt:
@yusukebe is the reason known why the node-server seems that 'slow'?

yusukebe:
Hi @kevinvdburgt 

Node.js Incoming/Outgoing Messages and Web Standards Request/Response are incompatible, and the translations cause overhead.

marcosrjjunior:
maybe there is something we can learn from this implementation?

- [node-http-handler](https://github.com/smithy-lang/smithy-typescript/blob/main/packages/node-http-handler/src/node-http-handler.ts#L194)
- [http request](https://github.com/smithy-lang/smithy-typescript/blob/main/packages/protocol-http/src/httpRequest.ts)

not sure if that helps, I'm just sharing

PR #301, opened 2026-02-10
perf: optimize request body reading, URL construction, and header building

mgcrea:
Hono's Node.js adapter was underperforming compared to h3, fastify, and koa — especially on body parsing. These changes bring Hono from 4th place to 2nd, virtually tied with h3. Addresses #76 

On my macOS (M1-Max)

### Before

| Framework | Average    | Ping      | Query     | Body      |
| --------- | ---------- | --------- | --------- | --------- |
| h3        | 51,963     | 58,842    | 50,739    | 46,310    |
| fastify   | 49,652     | 59,478    | 57,415    | 32,064    |
| koa       | 45,144     | 49,900    | 47,491    | 38,040    |
| **hono**  | **42,552** | **55,183**| **46,033**| **26,439**|
| express   | 42,408     | 43,229    | 40,298    | 43,698    |

### After

| Framework | Average    | Ping      | Query     | Body      |
| --------- | ---------- | --------- | --------- | --------- |
| h3        | 53,263     | 59,200    | 52,123    | 48,468    |
| **hono**  | **53,102** | **59,697**| **52,883**| **46,726**|
| fastify   | 48,637     | 58,192    | 56,054    | 31,666    |
| koa       | 45,190     | 50,062    | 47,347    | 38,161    |
| express   | 41,863     | 41,848    | 39,773    | 43,969    |

### Improvement

| Benchmark | Before | After  | Change    |
| --------- | ------ | ------ | --------- |
| Ping      | 55,183 | 59,697 | **+8.2%** |
| Query     | 46,033 | 52,883 | **+14.9%**|
| Body      | 26,439 | 46,726 | **+76.7%**|
| Average   | 42,552 | 53,102 | **+24.8%**|

Benchmarked with [bun-http-framework-benchmark](https://github.com/SaltyAom/bun-http-framework-benchmark) using `bombardier` at 500 concurrent connections for 10s per test.

### 1. Direct body reading from `IncomingMessage` (`request.ts`)

The biggest bottleneck was body parsing. When `req.text()` or `req.json()` was called on the pseudo-Request, it triggered a long chain of allocations:

```
getRequestCache() → new AbortController() → newHeadersFromIncoming() → new Headers()
→ new Request(url, { body: new ReadableStream(Readable.toWeb(incoming)) }) → .text()
```

This created 5+ intermediate objects and piped the body through multiple stream conversion layers on every request.

The fix overrides `text()`, `json()`, `arrayBuffer()`, and `blob()` on the request prototype to read directly from the Node.js `IncomingMessage` using event-based I/O:

```typescript
incoming.on('data', (chunk) => chunks.push(chunk))
incoming.on('end', () => resolve(Buffer.concat(chunks)))
```

If `getRequestCache()` is later called (e.g. user accesses `req.signal`), the cached buffer is reused as the body source to avoid double-consumption.

Falls back to the original `getRequestCache()` path when the full `Request` object has already been materialized.

### 2. URL construction fast-path (`request.ts`)

`new URL()` was called for every request purely for validation and normalization. Since Hono's `getPath()` and `getQueryParam()` both do string-based parsing (no URL object needed), we skip `new URL()` for common requests:

- **Fast path**: validate host with a charCode loop, concatenate `${scheme}://${host}${path}`
- **Slow path**: fall back to `new URL()` only when `..` is present in the path (needed for path normalization)

### 3. `buildOutgoingHttpHeaders` optimization (`utils.ts`)

Added `headers.has('set-cookie')` check before iterating headers. In the common case (no cookies), this skips the per-header string comparison against `'set-cookie'`.

---

Thanks for the great lib :)

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -69,16 +69,22 @@ export const buildOutgoingHttpHeaders = (
     headers = new Headers(headers ?? undefined)
   }
 
-  const cookies = []
-  for (const [k, v] of headers) {
-    if (k === 'set-cookie') {
-      cookies.push(v)
-    } else {
+  if (headers.has('set-cookie')) {
+    const cookies = []
+    for (const [k, v] of headers) {
+      if (k === 'set-cookie') {
+        cookies.push(v)
+      } else {
+        res[k] = v
+      }
+    }
+    if (cookies.length > 0) {
+      res['set-cookie'] = cookies
+    }
+  } else {
+    for (const [k, v] of headers) {
       res[k] = v
     }
-  }
-  if (cookies.length > 0) {
-    res['set-cookie'] = cookies
   }
   res['content-type'] ??= 'text/plain; charset=UTF-8'
 


yusukebe:
Hey @mgcrea Thank you for the PR! This is awesome. We started to consider how to handle this. Please wait a bit.

usualoma:
Hi @mgcrea 
Sorry for the late reply on this!

Thanks so much for this work — we really appreciate the thorough analysis and the clear breakdown of each optimization.

All three changes look effective, and we'd like to adopt all of them. Here are our thoughts on each:

**3. `buildOutgoingHttpHeaders` optimization**
This one looks good as-is. We're happy to accept it directly.

**1. Direct body reading from `IncomingMessage`**
This is the biggest win, but it also carries the most risk. We need to be very careful about edge cases — for example, handling client-side aborts mid-stream, and ensuring robust error handling throughout the event-based reading path. Since any issues here would directly affect every request with a body, we want to add thorough exception handling before shipping this.

**2. URL construction fast-path**
We discussed a similar area in #126 — any divergence from `new URL(url).href` behavior can introduce security vulnerabilities (e.g., path traversal bypasses). The `..` check is a good start, but we'll need to carefully tighten the conditions to ensure the fast path never produces a URL that differs from what `new URL()` would return in security-relevant ways.

**Our plan:**
Since all three optimizations need varying degrees of hardening, we won't merge this directly into `main`. Instead, we'd like to merge this into a dedicated performance-improvement branch, where we can iterate on the necessary adjustments for (1) and (2). We'll be working on those follow-up changes, and we'd love your input when we post updates — your feedback would be very welcome.

Again, thanks for the great contribution! 🙏

usualoma:
We've created a `perf/request-v2` branch to work on the optimizations discussed above. We'll merge this PR into that branch now.

Thanks again for the great contribution — we'll keep you posted as we iterate on the hardening work! 🙏

PR #302, opened 2026-02-11
build: migrate from tsup to tsdown

tommy-ish:
I migrated build tool from tsup to tsdown.

I ran `npx tsdown-migrate` according to tsdown's [Migration Guide](https://tsdown.dev/guide/migrate-from-tsup#migration-guide).

I modified the tsdown configuration to align with the paths defined in package.json.

I added a separate ci-build job running on Node.js 24, since tsdown (Rolldown) requires node:util#styleText which is not available in Node.js 18.


rafaell-lycan:
Does it impact the output size? 🤔

yusukebe:
The output size will be reduced significantly!

yusukebe:
@tommy-ish 

Sorry for being late. Adding `ci-build` is good! I'll merge this into the v2 branch for the next major release. Thanks!

PR #305, opened 2026-02-15
chore: add benchmark for comparing with npm and local (dev)

yusukebe:
This PR introduces a benchmark project to compare the Node.js Adapter on npm and the local version. With this, you can benchmark which local developer's version is faster. Very useful for performance tuning.

The `app.js` file does not use Hono; it uses a native fetch-based app. This fills the gap in frameworks.

The code is almost written by Claude Code, but it works well.

### DIFF
--- a/benchmarks/fetch/.gitignore
+++ b/benchmarks/fetch/.gitignore
@@ -1,0 +1,2 @@
+package-lock.json
+node_modules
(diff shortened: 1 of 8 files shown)


yusukebe:
Hey @usualoma !

I believe this is useful for us. The code is not critical and seems to be working fine. I'll merge immediately, but the problems may be nothing. Please use this for your development.

usualoma:
Thank you!

PR #309, opened 2026-03-01
fix: do not overwrite Content-Length in the fast path pattern if Content-Length already exists.

### LINKED ISSUE #308: [Bug] Duplicate `Content-Length` Header when Returning Response with Lowercase `content-length` Header
Hi,

It seems that Hono node-server is injecting an extra `Content-Length` header to `Response` object with a lowercase `content-length` Header. Reproduction example:

```js
import { serve } from '@hono/node-server';
import { Hono } from 'hono';

const app = new Hono();

app.get('/', (c) => {
  return new Response('Hello, Hono!', {
    headers: { 'Content-Type': 'text/plain', 'content-length': '12' },
  });
});

serve(
  {
    fetch: app.fetch,
    port: 3000,
  },
  (info) => {
    console.log(`Server is running on http://localhost:${info.port}`);
  },
);
```
Then run curl to test:
```bash
$ curl http://localhost:3000/ -v
* Host localhost:3000 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:3000...
* Connected to localhost (::1) port 3000
> GET / HTTP/1.1
> Host: localhost:3000
> User-Agent: curl/8.7.1
> Accept: */*
> 
* Request completely sent off
< HTTP/1.1 200 OK
< Content-Type: text/plain
< content-length: 12
< Content-Length: 12
< Date: Sun, 22 Feb 2026 02:30:28 GMT
< Connection: keep-alive
< Keep-Alive: timeout=5
< 
* Connection #0 to host localhost left intact
Hello, Hono!%
```
Impact:
Minor inconvenience as it breaks a API client that I use, but browsers are handling it fine.

Cause:
`listener.js` is injecting the `Content-Type` header [here](https://github.com/honojs/node-server/blob/cc05c48d8eb34007d5d5ff3861ea554721387041/src/listener.ts#L79), without first checking if the `content-type` header exists (and the case-insensitivity nature of headers is making it harder to check, perhaps the global `Header` object is helpful, but constructing the headers again might slow things down)

Why am I setting the `content-length` header?
I use Hono to forward a response generated by svelte's node server, it contains the header by default.

usualoma:
fixes #308

This PR contains two fixes related to the lightweight Response fast path.

### For 7cfd1e37fac78b1f0c33760d9c43503e135e3bd8

* It now checks whether `content-length` is already present before setting Content-Length.
* It is particularly optimized for cases where no header is specified at all. (for example, `new Response('text')`)
* The literal `{ 'content-type': 'text/plain; charset=UTF-8' }` now appears in two places. This duplication is intentional: we prioritize performance in the fast path over deduplication here.

### For 46ebb313c62070f39537a01b834285e7e21487dc

* Tuple-style headers ([string, string][]) were not handled in responseViaCache before.
* Creating a `new Headers(...)` instance is relatively expensive, but this header style is expected to be very rare in practice, so we accept this tradeoff to keep the implementation simple.

### DIFF
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -77,6 +77,10 @@ const responseViaCache = async (
   } else if (header instanceof Headers) {
     hasContentLength = header.has('content-length')
     header = buildOutgoingHttpHeaders(header)
+  } else if (Array.isArray(header)) {
+    const headerObj = new Headers(header)
+    hasContentLength = headerObj.has('content-length')
+    header = buildOutgoingHttpHeaders(headerObj)
   } else {
     for (const key in header) {
       if (key.length === 14 && key.toLowerCase() === 'content-length') {
--- a/src/response.ts
+++ b/src/response.ts
@@ -10,7 +10,7 @@ export const cacheKey = Symbol('cache')
 export type InternalCache = [
   number,
   string | ReadableStream,
-  Record<string, string> | Headers | OutgoingHttpHeaders | undefined,
+  Record<string, string> | [string, string][] | Headers | OutgoingHttpHeaders | undefined,
 ]
 interface LightResponse {
   [responseCache]?: globalThis.Response
--- a/test/server.test.ts
+++ b/test/server.test.ts
@@ -254,6 +254,25 @@ describe('various response body types', () => {
         })
         return response
       })
+      app.get('/text-with-content-length-array', () => {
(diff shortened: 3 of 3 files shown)


usualoma:
@yusukebe
Would you please review this?

yusukebe:
@usualoma This is a good PR! Thanks.

PR #317, opened 2026-03-12
feat: end support Node.js v18

yusukebe:
Node.js v18 reached EOL. In v2, we removed the support for it.


yusukebe:
Hi @usualoma 

I think this is good. Can you review this?

usualoma:
@yusukebe Thank you 👍

PR #311, opened 2026-03-01
perf: improve body read performance

usualoma:
This PR applies fixes to maintain compatibility with web standards for the request body optimization merged in #301.

It seems we can implement this without significantly reducing the performance gains, but it will likely require a considerable amount of code. Maintenance costs also look substantial.

```
============================================================
BENCHMARK RESULTS
============================================================

| Benchmark         | npm            | dev            | Difference  |
| ----------------- | -------------- | -------------- | ----------- |
| Average           | 50,527.59      | 61,093.99      | +20.91%     |
| Ping (GET /)      | 65,812.06      | 70,197.87      | +6.66%      |
| Query (GET /id)   | 64,671.18      | 69,099.75      | +6.85%      |
| Body (POST /json) | 21,099.52      | 43,984.34      | +108.46%    |
```

PR #318, opened 2026-03-12
fix(request): return error object instead of throwing

usualoma:
Fixing the error introduced in #311


usualoma:
@yusukebe
Would you please review this?

yusukebe:
@usualoma Thank you for the quick response!

PR #323, opened 2026-03-13
fix: only build exported entrypoints

BlankParticle:
fix for https://github.com/honojs/node-server/pull/316#discussion_r2929874647


yusukebe:
@BlankParticle Nice!

PR #328, opened 2026-03-20
feat: first class support for websockets

### LINKED ISSUE #307: Add first class WebSocket support to `@hono/node-server`
Currently for WebSocket support in node we need to install `@honojs/node-ws` middleware along with `@honojs/node-server`, then create some helper functions with `createNodeWebSocket` and pass them around.
So a basic echo websocket server looks like this,
```ts
import { createNodeWebSocket } from "@hono/node-ws";
import { Hono } from "hono";

const app = new Hono();

const { injectWebSocket, upgradeWebSocket } = createNodeWebSocket({ app });

app.get(
  "/ws",
  upgradeWebSocket((c) => ({
    onMessage(event, ws) {
      ws.send(event.data);
    },
  })),
);

const server = serve(app);
injectWebSocket(server);
```
this is awkward because you need to create an app, then use that to create helper and then chain on the app again to setup helpers, and you also need to remember to call `injectWebSocket`. I personally had experiences where I didn't create the helpers properly, or had multiple routers but did use the correct one or forgot to inject and debug why my ws server is not working.

on the other side, `hono/bun` has first class support for WebSockets because bun makes it really easy to do so. You import `upgradeWebSocket` from `hono/bun` anywhere, and just need to pass in `websocket` from `hono/bun` to `Bun.serve`

thus, I wanted to make WebSockets a first class supported feature in `@hono/node-server`, I have made a branch on my fork exploring this and is currently using it.

```ts
// Using my fork
import { serve, upgradeWebSocket } from '@blankparticle/hono-node-server'
import { Hono } from 'hono'

const app = new Hono()

app.get(
  '/ws',
  upgradeWebSocket(() => ({
    onMessage(event, ws) {
      ws.send(event.data)
    },
  }))
)

serve({
  fetch: app.fetch,
  websocket: true, // opt into using websockets
})
```

npm: https://www.npmjs.com/package/@blankparticle/hono-node-server
github: https://github.com/honojs/node-server/compare/main...BlankParticle:node-server:blank/fork

this makes the experience same as `hono/bun`, adds `ws` as a dependency to `@hono/node-server`, I ran the benchmarks and I see a `±2%` difference when WebSockets is enabled, which can be taken as measurement error on my macbook.

If this is something we can add to `@hono/node-server`, I will be happy to make proper PRs with my changes, so we can add the first class WebSocket support, and deprecate `@hono/node-ws` and update the docs

Thanks!

BlankParticle:
fixes #307

There was some discussion around how we should do this, but it died down
I landed on a design that doesn't add any extra runtime dependency, the user needs to provide the websocket server themselves, which is a good compromise.

I am opening this PR so we can do discussion around this and easily reason about this, 
~~this is not the final design and I will update this PR accordingly to the reviews/comments to better fit the goal~~

code was ported and adapted from `@hono/node-ws`


yusukebe:
Hey @BlankParticle 

Sorry for being late. I think the direction of this PR is good! So, can you finalize this, or is this ready for review?

I'd like to include this change to the next major version v2: https://github.com/honojs/node-server/pull/316

BlankParticle:
This version is ready for review where you need to provide your own websocket server.
I was wondering if you have considered having `ws` as peer dependency as @nakasyou said would be fine in #307 
else this PR is good to go

yusukebe:
@BlankParticle

Simple question. If you add `ws` as `peerDependencies`, the user should install `ws` even though they don't use WebSocket? If so, adding `ws` to `peerDependencies` is not necessary; only adding docs is better, I think.

BlankParticle:
this design works then, users will need to pass in their own server,
this is also better for use cases like if you have a `ws` compatible library or you want to pass in an existing server like in a vite plugin

this PR is ready to go

yusukebe:
@BlankParticle Okay!

Hey @nakasyou can you review this? (We don't need to add `ws` to `peerDependencies`).

nakasyou (approved):
It looks great!

yusukebe (approved):
LGTM!

yusukebe:
@BlankParticle Looks good! Thank you!

LankyMoose:
thank you thank you thank you!

PR #333, opened 2026-04-03
v2: perf(response,listener): Response fast-paths and responseViaCache improvements

GavinMeierSonos:
## Summary

**Response:**
- Extend `InternalCache` to support `null` body (redirects, 204s)
- Override `Response.json()` to return a `LightweightResponse` so the listener fast-path (`cacheKey` check) is hit instead of falling through to `ReadableStream` reading
- Override `Response.redirect()` with the same approach; `null` body avoids a spurious `content-type: text/plain` header on redirects

**Listener:**
- `responseViaCache` no-header fast-path: build the final header object in one shot per body type (`string`, `Uint8Array`, `Blob`, stream, `null`) to avoid V8 hidden-class transitions from mutating a single-key object
- Fix `null`-body responses (redirects, 204s) getting a spurious `content-type: text/plain` header
- Extract `makeCloseHandler` as a module-level factory so the close-listener closure is only allocated for async responses, not every request
- Replace `incomingEnded` boolean tracking with `incoming.readableEnded`

## Changes

- `src/response.ts`
- `src/listener.ts`
- `test/response.test.ts`
- `test/listener.test.ts`

PR #342, opened 2026-04-13
fix: ensure close handler is attached for Blob/ReadableStream cacheable responses

usualoma:
The previous fast-path check (introduced in #333) (`cacheKey in res`) did not distinguish body types, allowing `Blob` and `ReadableStream` responses to skip close handler registration. This meant client disconnects during async body processing would not trigger `AbortSignal`.

Restrict the fast path to body types that complete synchronously (null, string, Uint8Array) via `isImmediateCacheableResponse`. All other cacheable responses now go through `ensureCloseHandler` for proper abort detection.

### DIFF
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -110,6 +110,16 @@ const makeCloseHandler =
     }
   }
 
+const isImmediateCacheableResponse = (res: Response): boolean => {
+  if (!(cacheKey in res)) {
+    return false
+  }
+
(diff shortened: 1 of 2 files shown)


usualoma:
Hi @yusukebe
Would you mind reviewing this?

yusukebe (approved):
LGTM!

yusukebe:
@usualoma 

Thanks! This should be fixed. Looks good.

PR #343, opened 2026-04-13
fix: improve Response.json() and Response.redirect() spec compliance and efficiency

usualoma:
The optimizations for `Response.json()` and `Response.redirect()` introduced in #333 cause behavior discrepancies with native `Response` objects in cases such as those discussed at https://github.com/honojs/hono/issues/2343.

While excluding the `Response.json()` optimization is under consideration, if it is included as is, I believe we should ensure that there are no behavioral discrepancies with the native `Response` object, as proposed in this pull request.

### What this pull request includes

- Throw TypeError for non-JSON-serializable data in Response.json()
- Add URL validation with fast-path for common ASCII URLs in Response.redirect()
- Remove redundant type assertions in Response.json()

PR #344, opened 2026-04-13
feat: Remove Response.json() and Response.redirect() overrides

usualoma:
If we want to keep the overrides for `Response.json()` and `Response.redirect()`, I think we should implement the fix from #343, but I also think it’s a reasonable decision not to include this optimization in v2.


usualoma:
Hi @yusukebe
I’m fine with either #343 or #344.

PR #346, opened 2026-04-29
fix: forward Hono response headers during WebSocket upgrade

gentamura:
## Summary

Forward headers attached to the `Response` returned by the Hono app to the WebSocket upgrade response, so headers set by middleware (e.g. `Set-Cookie`, custom auth headers, `WWW-Authenticate` on reject) are no longer dropped during the handshake.

Previously, in `src/websocket.ts` the `Response` returned from `fetchCallback(...)` was inspected only for `status` — `response.headers` was discarded on both the successful and rejected paths.

## Changes

- On a successful upgrade, response headers are appended via `wss.on('headers', ...)` (the official [`ws`](https://github.com/websockets/ws) API for injecting handshake headers). The listener is removed in `finally`; `headers` is emitted synchronously inside `handleUpgrade`, so it cannot leak across concurrent upgrades on the shared `wss`.
- On a rejected upgrade, response headers are written into the manual `socket.end(...)` HTTP response produced by `rejectUpgradeRequest`.
- Skip hop-by-hop headers per [RFC 9110 Section 7.6.1](https://www.rfc-editor.org/rfc/rfc9110.html#name-connection) (`connection`, `keep-alive`, `proxy-authenticate`, `proxy-authorization`, `te`, `trailer`, `transfer-encoding`, `upgrade`), framing (`content-length`), and WebSocket handshake headers managed by `ws` (`sec-websocket-accept`, `sec-websocket-extensions`, `sec-websocket-protocol`) to avoid corrupting the handshake.

## Tests

Two new tests in `test/websocket.test.ts`:

- A custom response header set by middleware reaches the client on a successful upgrade.
- A custom response header is preserved on a rejected (401) upgrade response.

## Verification

- `bun run test --run test/websocket.test.ts` (5/5 passing)
- `bun run format`
- `bun run lint` (pre-existing warning in `listener.ts` only, unrelated to this change)
- `bun run build`

## Context

A companion PR is open against `honojs/middleware` for the same bug in `@hono/node-ws`: honojs/middleware#1873. Since `@hono/node-ws` is being deprecated in favor of `@hono/node-server`'s built-in `upgradeWebSocket` (per honojs/middleware#1862), porting the fix here ensures the bug does not survive the migration.

### DIFF
--- a/src/websocket.ts
+++ b/src/websocket.ts
@@ -57,11 +57,45 @@ type UpgradeWebSocketOptions = {
   onError: (err: unknown) => void
 }
 
-const rejectUpgradeRequest = (socket: Duplex, status: number) => {
+// Hop-by-hop headers per RFC 9110 Section 7.6.1
+// (https://www.rfc-editor.org/rfc/rfc9110.html#name-connection) plus
+// `keep-alive` (commonly treated as hop-by-hop by HTTP implementations) and
+// WebSocket handshake headers managed by `ws` itself. These must not be
+// forwarded onto the upgrade response or the handshake will be corrupted.
+const responseHeadersToSkip = new Set([
+  'connection',
+  'content-length',
+  'keep-alive',
+  'proxy-authenticate',
+  'proxy-authorization',
+  'te',
+  'trailer',
+  'transfer-encoding',
+  'upgrade',
+  'sec-websocket-accept',
+  'sec-websocket-extensions',
+  'sec-websocket-protocol',
+])
+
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@gentamura 

Thanks! Good PR. I'll merge and release a new version now.

Regarding https://github.com/honojs/middleware/pull/1873, I think we can close it as the `@hono/node-ws` is already deprecated.

PR #350, opened 2026-05-06
fix(serve-static): stop using file birthtime for Date header

### LINKED ISSUE #348: serveStatic middleware sets the Date header to the file creation time, which is invalid for HTTP
The Date header should always be set to the message time. Setting it to the file creation time is invalid semantics. This might even be overwritten by the server, so this bug might never have appeared in practice.

usualoma:
fixes #348

## Changes

- Remove `Date: stats.birthtime` from range responses
  - The `Date` header should represent the time when the response message is generated, so using the file birth time was incorrect.
  - Node.js automatically adds the `Date` header, so `serveStatic` does not need to set it explicitly.
- Add `Last-Modified: stats.mtime` for successful static file responses
  - Returning the file `mtime` in the `Last-Modified` header is standard behavior for most web servers.
  - For precompressed files (`.gz`, `.br`, `.zst`), this uses the selected precompressed file's `mtime`. This can differ from the original content file's modification time, but that seems acceptable because it describes the actual representation being served.

### DIFF
--- a/src/serve-static.ts
+++ b/src/serve-static.ts
@@ -177,7 +177,6 @@ export const serveStatic = <E extends Env = any>(
       result = c.body(createStreamBody(createReadStream(path)), 200)
     } else {
       c.header('Accept-Ranges', 'bytes')
-      c.header('Date', stats.birthtime.toUTCString())
 
       const parts = range.replace(/bytes=/, '').split('-', 2)
       const start = parseInt(parts[0], 10) || 0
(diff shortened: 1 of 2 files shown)


usualoma:
Hi @yusukebe,

Would you mind reviewing this?

yusukebe (approved):
LGTM!

yusukebe:
@usualoma 

Thank you! This should be fixed. Looks good!

PR #352, opened 2026-05-07
chore(ci): update GitHub Actions versions

BlankParticle:
Node 20 actions will be deprecated on September 19, 2026. 
https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

Updating the workflows to `v6`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,6 @@ on:
 
 permissions:
   contents: read
-  pull-requests: read
 
 jobs:
   ci:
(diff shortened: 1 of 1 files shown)


lppedd:
Could we pin all actions by SHA?

.github/workflows/ci.yml
```diff
@@ -7,7 +7,6 @@ on:
 
 permissions:
   contents: read
-  pull-requests: read
```
yusukebe, line 10: This is not necessary.

yusukebe:
@BlankParticle Thanks! As @lppedd said, we should pin them with SHAs. I've updated myself.

PR #356, opened 2026-05-17
docs: Align the `ServeStaticOptions` comment with the current spec

### LINKED ISSUE #355: JSDoc for `ServeStaticOptions` seems not updated
https://github.com/honojs/node-server/pull/257 introduced support for absolute path in `serveStatic`, but the comments on type `ServeStaticOptions` still indicates that it does not support absolute paths:

```ts
export type ServeStaticOptions<E extends Env = Env> = {
  /**
   * Root path, relative to current working directory from which the app was started. Absolute paths are not supported.
   */
  root?: string
  // ...
}
```

https://github.com/honojs/node-server/blob/4aa0650744858993d1770d64aa4826b436304bfd/src/serve-static.ts#L10

This can be confusing for users.

kakkokari-gtyih:
## What
Modified the JSDoc comment on `ServeStaticOptions.root` to current spec (relative/absolute paths are both supported).

## Why

Fix #355

### DIFF
--- a/src/serve-static.ts
+++ b/src/serve-static.ts
@@ -7,7 +7,7 @@ import { createStreamBody } from './utils/stream'
 
 export type ServeStaticOptions<E extends Env = Env> = {
   /**
-   * Root path, relative to current working directory from which the app was started. Absolute paths are not supported.
+   * Root path. Relative path is based on current working directory from which the app was started.
    */
   root?: string
   path?: string


yusukebe (approved):
LGTM!

yusukebe:
@kakkokari-gtyih 

Thank you for the quick fix. Looks good!

PR #357, opened 2026-05-17
fix: preserve headers mutated after raw Response construction

### LINKED ISSUE #304: Header missing when returning raw response
It has been confirmed that this issues does not occur in the bun environment or the AWS Lambda environment, but only when using hono/node-server.

This issue can be reproduced when using the `cors` or `compress` middleware. 

```ts
import { Hono } from 'hono';
import { compress } from 'hono/compress';

const app = new Hono();

app.use(compress());

app.post('/test', async c => {
  const res = new Response('hello', { status: 200, headers: { 'Content-Type': 'application/json' } });
  res.headers.append('Set-Cookie', 'session=abc; Path=/; HttpOnly');
  return res;
});

export default app;
```

It seems the raw response header was lost when the middleware modified the header.

---

### Claude Code

When cloning a Response2 instance via new Response(body, existingResponse), headers that were appended after
  construction (e.g. Set-Cookie via response.headers.append()) are silently dropped.

  Root cause:

  In the Response2 constructor, when init is another Response2 instance and no responseCache exists, headers are read
  from init.#init.headers (the original ResponseInit passed at construction time) instead of init.headers (the current
  headers via the getter, which includes all subsequent mutations):

```ts
  // current (buggy)
  if (init instanceof _Response) {
    const cachedGlobalResponse = init[responseCache];
    if (cachedGlobalResponse) {
      this.#init = cachedGlobalResponse;
      this[getResponseCache]();
      return;
    } else {
      this.#init = init.#init;
      headers = new Headers(init.#init.headers); // ← reads stale, original headers
    }
  }
```

  Response2 stores headers in two places:
  - this[cacheKey][2] — the live Headers object returned by the headers getter, reflecting all mutations (.append(),
  .set(), .delete())
  - this.#init.headers — the original ResponseInit.headers passed to the constructor, never updated

  The clone reads from #init.headers, so any header added after construction is lost.

  Reproduction:

```ts
  const res = new Response('hello', { status: 200, headers: { 'Content-Type': 'application/json' } });
  res.headers.append('Set-Cookie', 'session=abc; Path=/; HttpOnly');

  const cloned = new Response(res.body, res);
  console.log(cloned.headers.get('set-cookie')); // null — should be "session=abc; Path=/; HttpOnly"
```

  This does not reproduce with the native Response (e.g. in Lambda / Workers runtimes), only with @hono/node-server's
  Response2 override.

  Impact:

  Any Hono middleware that clones the response after finalization (e.g. cors() calling c.header() which triggers new
  Response(this.#res.body, this.#res)) will lose Set-Cookie headers set by route handlers via raw Response objects.
  This breaks cookie-based authentication in local Node.js development while working fine in production on
  Workers/Lambda.

 Fix:

```ts
    } else {
      this.#init = init.#init;
  -   headers = new Headers(init.#init.headers);
  +   headers = new Headers(init.headers);
    }
```

  init.headers goes through the getter, which returns the live Headers from cacheKey[2] containing all mutations. new
  Headers(...) still creates a new independent copy, so parent and child do not share the same object.

abdulmunimjemal:
Fixes #304.

## Problem

When a Hono handler returns `new Response(body, init)` and appends headers after construction — e.g. `res.headers.append('Set-Cookie', ...)` — the appended headers are silently dropped once middleware such as `cors` or `compress` clones the response. The bug only manifests on `@hono/node-server`; the same code works correctly on Bun, Cloudflare Workers, and AWS Lambda.

Minimal repro from the issue:

```ts
import { Hono } from 'hono'
import { compress } from 'hono/compress'

const app = new Hono()
app.use(compress())

app.post('/test', async (c) => {
  const res = new Response('hello', { status: 200, headers: { 'Content-Type': 'text/plain' } })
  res.headers.append('Set-Cookie', 'session=abc; Path=/; HttpOnly')
  return res
})
```

The `Set-Cookie` header never reaches the client.

## Root cause

`src/response.ts` defines a lightweight `Response2` that stores headers in two places:

- `cacheKey[2]` — the live `Headers` object exposed by the `headers` getter, reflecting every `.append()` / `.set()` / `.delete()` mutation.
- `this.#init.headers` — the original `ResponseInit.headers` argument, never updated.

Two code paths read from `#init.headers` instead of the live headers:

1. **`getResponseCache`** built the underlying `GlobalResponse` from `this.#init`. Any access to a delegated property (`.body`, `.text()`, …) — which middleware does when streaming a raw body — froze the stale init headers and lost subsequent mutations.
2. **The constructor**, when cloning via `new Response(body, init)` where `init` is another `Response2` without a materialized `responseCache`, read `init.#init.headers` instead of `init.headers` (the getter that returns the live `cacheKey[2]`).

## Fix

Both spots now consult the live `Headers` instance on `cacheKey[2]` when present, so post-construction header mutations survive cloning. The fix is local to `src/response.ts`; no public API changes.

```diff
   [getResponseCache](): globalThis.Response {
+    const cache = (this as LightResponse)[cacheKey]
+    const liveHeaders = cache && cache[2] instanceof Headers ? cache[2] : undefined
     delete (this as LightResponse)[cacheKey]
-    return ((this as LightResponse)[responseCache] ||= new GlobalResponse(this.#body, this.#init))
+    return ((this as LightResponse)[responseCache] ||= new GlobalResponse(
+      this.#body,
+      liveHeaders ? { ...this.#init, headers: liveHeaders } : this.#init
+    ))
   }
```

```diff
       } else {
         this.#init = init.#init
-        headers = new Headers((init.#init as ResponseInit).headers)
+        headers = new Headers(init.headers)
       }
```

`new Headers(...)` still produces an independent copy, so parent and child do not share the same object — the existing "Should not lose header data" test continues to pass.

## Tests

Two regression tests added:

- `test/response.test.ts` — unit test covering both clone patterns:
  1. `new Response('body', parent)` before `getResponseCache` is triggered (exercises the constructor fix).
  2. `new Response(parent.body, parent)` after `.body` materializes the GlobalResponse (exercises the `getResponseCache` fix).
- `test/server.test.ts` — end-to-end test that mirrors the issue: a middleware re-wraps `c.res` via `new Response(c.res.body, c.res)`, and the handler returns a raw `Response` with an appended `Set-Cookie`. Asserts that the `Set-Cookie` survives.

Both tests fail on `main` and pass with the fix applied. The full suite stays green: **343 passed / 343** (12 files).

### DIFF
--- a/src/response.ts
+++ b/src/response.ts
@@ -25,8 +25,16 @@ export class Response {
   #init?: ResponseInit;
 
   [getResponseCache](): globalThis.Response {
+    // If `cacheKey` has been populated with a live `Headers` instance, the
+    // user (or middleware) may have mutated it after construction. Use those
+    // headers so the GlobalResponse reflects the current state.
+    const cache = (this as LightResponse)[cacheKey]
+    const liveHeaders = cache && cache[2] instanceof Headers ? cache[2] : undefined
     delete (this as LightResponse)[cacheKey]
-    return ((this as LightResponse)[responseCache] ||= new GlobalResponse(this.#body, this.#init))
+    return ((this as LightResponse)[responseCache] ||= new GlobalResponse(
+      this.#body,
+      liveHeaders ? { ...this.#init, headers: liveHeaders } : this.#init
+    ))
   }
 
   constructor(body?: BodyInit | null, init?: ResponseInit) {
@@ -41,8 +49,11 @@ export class Response {
         return
       } else {
         this.#init = init.#init
-        // clone headers to avoid sharing the same object between parent and child
-        headers = new Headers((init.#init as ResponseInit).headers)
+        // Read headers via the live getter so mutations made on `init` after
+        // construction (e.g. `init.headers.append('Set-Cookie', ...)`) are
+        // preserved on the clone. `new Headers(...)` still produces an
+        // independent copy so parent and child do not share the same object.
+        headers = new Headers(init.headers)
       }
     } else {
       this.#init = init
--- a/test/response.test.ts
+++ b/test/response.test.ts
@@ -89,6 +89,34 @@ describe('Response', () => {
     expect(childResponse.headers.get('content-type')).toEqual('application/json')
   })
 
+  it('Should preserve headers mutated after construction when cloned via new Response(body, init)', () => {
+    // Regression test for https://github.com/honojs/node-server/issues/304.
+    // Headers appended (or set/deleted) after construction must be visible on
+    // a clone built with `new Response(body, parent)` — which is the pattern
+    // used by middleware such as `cors` and `compress`.
+    const parentResponse = new Response('hello', {
+      status: 200,
+      headers: { 'content-type': 'application/json' },
+    })
+    parentResponse.headers.append('set-cookie', 'session=abc; Path=/; HttpOnly')
+
+    // Pattern 1: clone before any `getResponseCache`-triggering access.
+    const childResponse = new Response('hello', parentResponse)
+    expect(childResponse.headers.get('set-cookie')).toEqual('session=abc; Path=/; HttpOnly')
+    expect(childResponse.headers.get('content-type')).toEqual('application/json')
+
+    // Pattern 2: clone after `.body` has materialized the GlobalResponse —
+    // this is what middleware does when streaming a raw Response body.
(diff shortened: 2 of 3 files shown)


usualoma:
Hi @abdulmunimjemal,
Thank you. LGTM!

yusukebe (approved):
LGTM!

PR #358, opened 2026-05-22
Declare public type dependencies

### LINKED ISSUE #353: Public declaration files reference undeclared dependency `ws`

Baijack-star:
## Summary

- move `@types/ws` into `dependencies` because the published root declarations import from `ws`
- move `@types/node` into `dependencies` because the published declarations also expose Node built-in server/request types
- update `bun.lock` accordingly

## Verification

- Reproduced the current published package failure in a clean consumer project with only `typescript`, `hono`, and `@hono/node-server@2.0.3` installed:
  - `npx tsc -p tsconfig.json` reports `TS2307: Cannot find module 'ws' or its corresponding type declarations`, plus missing Node type declarations.
- Packed this branch locally and installed it into a clean consumer project with only `typescript`, `hono`, and the local `@hono/node-server` tarball installed:
  - `npx tsc -p tsconfig.json` passes.
- `bun run build` passes, including `publint`.
- `git diff --check` passes.

Note: I also tried `bun test --run`, but the local full suite fails/hangs in unrelated Bun runtime HTTP/WebSocket cases on this machine (for example `readWithoutBlocking > should return the body for large text` and several abort/WebSocket timeout cases). The packaging/type-surface repro above is scoped to this issue.

I used an AI coding assistant to prepare this patch.

Fixes #353

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -4,11 +4,13 @@
   "workspaces": {
     "": {
       "name": "@hono/node-server",
+      "dependencies": {
+        "@types/node": "^20.10.0",
+        "@types/ws": "^8.18.1",
+      },
       "devDependencies": {
         "@hono/eslint-config": "^1.0.1",
-        "@types/node": "^20.10.0",
         "@types/supertest": "^2.0.12",
-        "@types/ws": "^8.18.1",
         "@whatwg-node/fetch": "^0.9.14",
         "eslint": "^9.10.0",
         "hono": "^4.12.8",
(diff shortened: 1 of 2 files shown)


alex-kinokon:
Don’t think it’s a great idea to pin the Node.js version in a direct dependency.

yusukebe:
This PR is not needed since https://github.com/honojs/node-server/pull/359 has been merged.

PR #359, opened 2026-05-23
fix: stub ws types to prevent them leaking in public types

### LINKED ISSUE #353: Public declaration files reference undeclared dependency `ws`

BlankParticle:
fixes #353


yusukebe (approved):
LGTM!

yusukebe:
@BlankParticle 

Thank you! Looks good.

PR #363, opened 2026-06-22
fix: preserve status and statusText when cloning a Response with liveheaders

### LINKED ISSUE #362: Regression: lightweight Response materialization changes status from 302 to 200
## Summary

This looks like a regression from https://github.com/honojs/node-server/pull/357. That PR added a materialization path that spreads `this.#init`:

```ts
new GlobalResponse(this.#body, {
  ...this.#init,
  headers: liveHeaders,
})
```

This is safe when `this.#init` is a plain `ResponseInit`, but unsafe when it is a native `Response`. Native `Response` fields such as `status` and `statusText` are accessor properties, not own enumerable properties, so object spread drops them:

```js
const native = new Response(null, { status: 302 })
console.log({ ...native }) // {}
```

As a result, materialization can effectively call `new GlobalResponse(body, { headers })`, which defaults to status `200`.

With `overrideGlobalObjects: true` (the default), this can make `@hono/node-server`'s lightweight `Response` lose its status during materialization. This is not specific to redirects: any non-default status from the native `Response` can be ignored and replaced with `200`.

A response created from a native `Response`:

```js
const lightweight = new Response(nativeResponse.body, nativeResponse)
```

initially reports `302`, but after `.body` is read it can become `200`.

## Regression Range

Tested with Node.js `24.16.0` and `hono@4.10.6`:

| `@hono/node-server` | Result |
| --- | --- |
| `2.0.2` | OK: `302` stays `302` |
| `2.0.3` | Broken: `302` becomes `200` |
| `2.0.4` | Broken: `302` becomes `200` |

## Reproduction

```js
const NativeResponse = globalThis.Response
const { serve } = require('@hono/node-server')

const server = serve(
  {
    fetch: () => new Response('ok'),
    hostname: '127.0.0.1',
    port: 0,
  },
  () => {}
)
server.close()

const nativeRedirect = new NativeResponse(null, {
  status: 302,
  statusText: '',
  headers: {
    Location: 'https://example.com/',
    'Content-Length': '0',
  },
})

const lightweightRedirect = new Response(nativeRedirect.body, nativeRedirect)

console.log('initial:', lightweightRedirect.status)

const rebuilt = new Response(lightweightRedirect.body, {
  status: lightweightRedirect.status,
  statusText: lightweightRedirect.statusText,
  headers: lightweightRedirect.headers,
})

console.log('after body read:', lightweightRedirect.status)
console.log('rebuilt:', rebuilt.status)
```

Actual on `2.0.3` / `2.0.4`:

```txt
initial: 302
after body read: 200
rebuilt: 200
```

Expected:

```txt
initial: 302
after body read: 302
rebuilt: 302
```

## Suggested Fix

Do not spread a native `Response` as if it were a plain `ResponseInit`. Read metadata through accessors:

```ts
const init =
  this.#init instanceof GlobalResponse
    ? {
        status: this.#init.status,
        statusText: this.#init.statusText,
        headers: liveHeaders ?? this.#init.headers,
      }
    : liveHeaders
      ? { ...this.#init, headers: liveHeaders }
      : this.#init

return new GlobalResponse(this.#body, init)
```

Related:

- https://github.com/honojs/node-server/issues/304
- https://github.com/honojs/node-server/pull/357
- https://github.com/honojs/node-server/issues/321

usualoma:
fixes #362

The lightweight Response built its GlobalResponse via `{ ...this.#init }`, but `#init` may hold a Response instance whose `status`/`statusText` are prototype getters and are lost when spread. Read those fields explicitly so they survive alongside the live headers.

### DIFF
--- a/src/response.ts
+++ b/src/response.ts
@@ -33,7 +33,13 @@ export class Response {
     delete (this as LightResponse)[cacheKey]
     return ((this as LightResponse)[responseCache] ||= new GlobalResponse(
       this.#body,
-      liveHeaders ? { ...this.#init, headers: liveHeaders } : this.#init
+      liveHeaders
+        ? {
+            status: this.#init?.status,
+            statusText: this.#init?.statusText,
+            headers: liveHeaders,
+          }
+        : this.#init
     ))
   }
 
--- a/test/response.test.ts
+++ b/test/response.test.ts
@@ -89,32 +89,54 @@ describe('Response', () => {
     expect(childResponse.headers.get('content-type')).toEqual('application/json')
   })
 
-  it('Should preserve headers mutated after construction when cloned via new Response(body, init)', () => {
-    // Regression test for https://github.com/honojs/node-server/issues/304.
-    // Headers appended (or set/deleted) after construction must be visible on
-    // a clone built with `new Response(body, parent)` — which is the pattern
-    // used by middleware such as `cors` and `compress`.
+  it('Should preserve mutated headers when cloned before body access', () => {
     const parentResponse = new Response('hello', {
       status: 200,
       headers: { 'content-type': 'application/json' },
     })
     parentResponse.headers.append('set-cookie', 'session=abc; Path=/; HttpOnly')
 
-    // Pattern 1: clone before any `getResponseCache`-triggering access.
     const childResponse = new Response('hello', parentResponse)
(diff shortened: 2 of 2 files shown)


usualoma:
Hi @yusukebe,
Could you review it?

yusukebe (approved):
LGTM!

yusukebe:
@usualoma 

Indeed, this should be fixed. Thank you!

PR #366, opened 2026-07-01
fix(serve-static): serve precompressed files for application/octet-stream

### LINKED ISSUE #360: Please document the `COMPRESSIBLE_CONTENT_TYPE_REGEX`-related behavior
In the [serveStatic](https://github.com/honojs/node-server/blob/9e1cdee836baeb13f33e857a582b29546185c324/src/serve-static.ts#L21) middleware, there exists a check for compressible content types (in the `COMPRESSIBLE_CONTENT_TYPE_REGEX`)... if this check fails, then the compressed version of a file will never be served, even IF there does exist a compressed variant of that file!

While experimenting with using [compression for a large binary file in one of my web projects](https://moursy.org/funbox/autoconnections), I found after analysing the network requests that the large file was not being served as a compressed Brotli file like it should have been (thus wasting precious network bandwidth and user time)!

When I saw that other files, like the actual `.html` were being served compressed, I had a hunch something was awry in the `serveStatic` middleware. After peeking the source code, I found that my hunch was correct, and renaming file from `.bin` to `.txt` (although technically incorrect) made Hono start serving the compressed version of the large file!

Considering that this behaviour is undocumented and could lead to user confusion, I ask that you document this somewhere.

yusukebe:
Fixes #360

`serveStatic({ precompressed: true })` didn't serve precompressed variants for `application/octet-stream` files like `.bin`, even when a `.br`/`.zst`/`.gz` existed. Now it does.

### DIFF
--- a/src/serve-static.ts
+++ b/src/serve-static.ts
@@ -108,7 +108,12 @@ export const serveStatic = <E extends Env = any>(
     const mimeType = getMimeType(path)
     c.header('Content-Type', mimeType || 'application/octet-stream')
 
-    if (options.precompressed && (!mimeType || COMPRESSIBLE_CONTENT_TYPE_REGEX.test(mimeType))) {
+    if (
+      options.precompressed &&
+      (!mimeType ||
+        mimeType === 'application/octet-stream' ||
+        COMPRESSIBLE_CONTENT_TYPE_REGEX.test(mimeType))
+    ) {
       const acceptEncodingSet = new Set(
         c.req
           .header('Accept-Encoding')
--- a/test/assets/static-with-precompressed/hello.bin
+++ b/test/assets/static-with-precompressed/hello.bin
@@ -1,0 +1,1 @@
+Hello Not Compressed
--- a/test/assets/static-with-precompressed/hello.bin.br
+++ b/test/assets/static-with-precompressed/hello.bin.br
@@ -1,0 +1,1 @@
+Hello br Compressed
--- a/test/serve-static.test.ts
+++ b/test/serve-static.test.ts
@@ -277,6 +277,18 @@ describe('Serve Static Middleware', () => {
     expect(res.text).toBe('Hello Not Compressed')
   })
 
+  it('Should return a pre-compressed response for an octet-stream file - /static-with-precompressed/hello.bin', async () => {
+    const res = await request(server)
+      .get('/static-with-precompressed/hello.bin')
+      .set('Accept-Encoding', 'gzip, br, zstd')
+    expect(res.status).toBe(200)
+    expect(res.headers['content-type']).toBe('application/octet-stream')
+    expect(res.headers['content-length']).toBe('19')
+    expect(res.headers['content-encoding']).toBe('br')
+    expect(res.headers['vary']).toBe('Accept-Encoding')
+    expect(res.body.toString()).toBe('Hello br Compressed')
+  })
+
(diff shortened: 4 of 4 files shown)


yusukebe:
Hey @usualoma 

I created this PR to fix #360. I thought it was better to handle `application/octet-stream` rather than update the docs as #360 said. Can you review?

usualoma:
Hi @yusukebe,
Thank you.

By the way, is the reason you didn't embed it directly into `COMPRESSIBLE_CONTENT_TYPE_REGEX` here to avoid discrepancies with the `COMPRESSIBLE_CONTENT_TYPE_REGEX` in the Hono core?
https://github.com/honojs/hono/blob/b20d4225c9aa9e615717ca74e437852cc9fd5b94/src/utils/compress.ts#L9-L10

I understand the desire to distinguish this from the `COMPRESSIBLE_CONTENT_TYPE_REGEX` in the Hono core, since it’s also used by the “compress” middleware.

-----

However, I think there’s also the view that, in the case of `precompressed`, the `COMPRESSIBLE_CONTENT_TYPE_REGEX` check isn’t necessary.

I feel it might also be a good idea to allow passing 
```diff
diff --git i/src/serve-static.ts w/src/serve-static.ts
index 9daa4c8..bea3535 100644
--- i/src/serve-static.ts
+++ w/src/serve-static.ts
@@ -12,7 +12,7 @@ export type ServeStaticOptions<E extends Env = Env> = {
   root?: string
   path?: string
   index?: string // default is 'index.html'
-  precompressed?: boolean
+  precompressed?: boolean | ((options: { path: string; mimeType?: string }) => boolean)
   rewriteRequestPath?: (path: string, c: Context<E>) => string
   onFound?: (path: string, c: Context<E>) => void | Promise<void>
   onNotFound?: (path: string, c: Context<E>) => void | Promise<void>
@@ -65,6 +65,12 @@ export const serveStatic = <E extends Env = any>(
     console.error(`serveStatic: root path '${root}' is not found, are you sure it's correct?`)
   }
 
+  const precompressedFilter: (options: { path: string; mimeType?: string }) => boolean =
+    typeof options.precompressed === 'function'
+      ? options.precompressed
+      : ({ mimeType }) =>
+          options.precompressed && (!mimeType || COMPRESSIBLE_CONTENT_TYPE_REGEX.test(mimeType))
+
   return async (c, next) => {
     // Do nothing if Response is already set
     if (c.finalized) {
@@ -108,7 +114,7 @@ export const serveStatic = <E extends Env = any>(
     const mimeType = getMimeType(path)
     c.header('Content-Type', mimeType || 'application/octet-stream')
 
-    if (options.precompressed && (!mimeType || COMPRESSIBLE_CONTENT_TYPE_REGEX.test(mimeType))) {
+    if (precompressedFilter({ path, mimeType })) {
       const acceptEncodingSet = new Set(
         c.req
           .header('Accept-Encoding')
```

-----

I thought about something like that, but I think it’s perfectly fine to target `application/octet-stream` for now!

yusukebe:
@usualoma 

> By the way, is the reason you didn't embed it directly into `COMPRESSIBLE_CONTENT_TYPE_REGEX` here to avoid discrepancies with the `COMPRESSIBLE_CONTENT_TYPE_REGEX` in the Hono core?

Yes. Right!

I like your approach of passing a function to `precompressed`, but in this case, handling only `application/octet-stream` is okay.

Anyway, thank you!

PR #367, opened 2026-07-01
chore: migrate to pnpm

### LINKED ISSUE #365: [Discussion] Changing package manager
following my PR in honojs/middleware, https://github.com/honojs/middleware/pull/2006
I was looking into other hono repos if we should standardize the same package manager across all the repos, it would be weird for the `node-server` to use bun as a runtime in dev/testing, but looks like bun is used for package manager only. 

`honojs/{hono,honox,website}` uses bun, `honojs/{vite-plugins,starter}` uses yarn
Should we put more effort into standardizing all the repos or is it fine to keep these repos on bun/yarn as it is right now?
cc @yusukebe

BlankParticle:
follow up to https://github.com/honojs/middleware/pull/2006
fixes #365 

`bun.lock` can't be directly imported by pnpm, in bunfig `install.lockfile.print = "yarn"` will generate yarn.lock file, which was used to generate the pnpm lock file


README.md
```diff
@@ -41,16 +41,11 @@ It works on Node.js versions greater than 20.x.
...
+You can install it from the npm registry:
 
 ```sh
 npm install @hono/node-server
-```
-
-Or use `yarn`:
-
-```sh
-yarn add @hono/node-server
```
yusukebe, line 53: Moving `yarn` to `pnpm` in the README is a good idea!

yusukebe (approved):
LGTM!

yusukebe:
@BlankParticle 

Nice! Looks good to me. Thanks.

PR #371, opened 2026-07-10
fix(websocket): polyfill missing ErrorEvent global

otnc:
Node.js has no global `ErrorEvent` (unlike `Event`, `MessageEvent`, and `CloseEvent`,
which are already polyfilled/native). When the underlying `ws` socket emits an
`error` event, `upgradeWebSocket`'s handler does `new ErrorEvent('error', { error })`,
which throws a `ReferenceError` on every supported Node version. That error is caught
and passed to the internal `onError` fallback instead — so the user's `onError`
handler passed to `upgradeWebSocket` is never invoked, and the original error is lost.

This adds an `ErrorEvent` polyfill following the same pattern as the existing
`CloseEvent` one, and adds a test that reproduces the bug by emitting a real
`error` event on the raw socket and asserting the user's `onError` receives it.

### DIFF
--- a/src/websocket.ts
+++ b/src/websocket.ts
@@ -38,6 +38,49 @@ export const CloseEvent: typeof globalThis.CloseEvent =
     }
   }
 
+interface ErrorEventInit extends EventInit {
+  message?: string
+  filename?: string
+  lineno?: number
+  colno?: number
+  error?: unknown
+}
+
+/**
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@otnc 

Good catch! Looks good. Thank you!

PR #373, opened 2026-07-13
fix: don't reject with unusable body when full body was received

### LINKED ISSUE #370: v2: body read after client disconnect rejects with `TypeError: Body is unusable` even when the full body was received (regression from v1)
## What version of Hono are you using?

- `@hono/node-server` 2.0.8 (regression — 1.19.14 is not affected)
- `hono` 4.12.28
- Node.js 25.0.0 (also observed in CI on Node 24)

## What happened?

If a client sends a **complete** request body and then disconnects (aborts the socket) while the server is still doing async work, any subsequent body read (`c.req.text()` / `.json()` / `.arrayBuffer()`) rejects with `TypeError: Body is unusable`.

On v1.19.14, the identical scenario succeeds — the body was fully received and buffered by Node, and the read returns it.

This is easy to hit in real apps: any handler that does async work between routing and the body read (auth, context creation, an ORM call) has a window in which a client abort poisons the body. `@trpc/server`'s fetch adapter is a prominent example — it builds the request context before reading inputs, so Playwright e2e suites that navigate away mid-mutation produce intermittent phantom `BAD_REQUEST: Body is unusable` errors on the server.

Possibly related but distinct from honojs/hono#4031 (throws from undici's `consumeBody` via the ETag middleware's double-read, not from node-server's `readBodyDirect`) and honojs/hono#4259 (a genuine double-read across two `zValidator`s, fixed in hono core). This one needs no middleware and no second read — a single read after a client disconnect is enough.

## Minimal reproduction

`server.mjs`:

```js
import { serve } from "@hono/node-server"
import { Hono } from "hono"

const app = new Hono()
app.post("/echo", async (c) => {
  // Simulate async work between routing and the body read
  // (auth, context creation — what e.g. @trpc/server does)
  await new Promise((r) => setTimeout(r, 15))
  try {
    const body = await c.req.text()
    return c.json({ len: body.length })
  } catch (e) {
    console.error("READ FAILED:", e.constructor.name, e.message)
    return c.json({ error: e.message }, 400)
  }
})
serve({ fetch: app.fetch, port: 3996 }, () => console.log("up"))
```

`client.mjs`:

```js
import http from "node:http"

// Send the FULL body, then destroy the socket while the server
// is still in its async work, before it reads the body.
const body = JSON.stringify({ hello: "world" })
const req = http.request(
  {
    host: "localhost",
    port: 3996,
    path: "/echo",
    method: "POST",
    headers: {
      "content-type": "application/json",
      "content-length": Buffer.byteLength(body),
    },
  },
  () => {}
)
req.on("error", () => {})
req.end(body)
setTimeout(() => req.destroy(), 5)
setTimeout(() => process.exit(0), 100)
```

Run `node server.mjs`, then `node client.mjs` a few times.

**v2.0.8** — fails every time:

```
READ FAILED: TypeError Body is unusable
```

**v1.19.14** — same server/client, zero failures; the read returns the buffered body.

## Expected behavior

Either of:

1. The read succeeds — the full body arrived and is sitting in the `IncomingMessage` buffer (this is v1's behavior via `Readable.toWeb`), or
2. If the body genuinely cannot be delivered, an error that says what actually happened (client disconnect), not `Body is unusable`, which points developers at a nonexistent double-read in their own code.

## Root cause (as far as I can tell)

`readBodyDirect` in v2's lazy body-reading path rejects up front when `Readable.isDisturbed(incoming)` is true:

```js
const readBodyDirect = (request) => {
  ...
  if (Readable.isDisturbed(incoming)) return rejectBodyUnusable();
```

But `Readable.isDisturbed()` returns true when the stream **has errored**, not only when it has been read from. A client abort errors the socket/stream, so the check conflates "body already consumed" (the case `Body is unusable` is meant for) with "socket errored after the body fully arrived" (recoverable — the data is buffered).

A check on `incoming.readableDidRead` alone would distinguish the two; the errored-but-unread case could fall through to the existing `data`/`end`/`close` listener logic, which already produces the accurate `"Client connection prematurely closed."` error when the body really is incomplete.

Happy to provide more detail — we also reproduced this at load against a production-shaped app (Hono + `@hono/trpc-server`, 4-parallel-worker Playwright CI): ~40 phantom errors per few thousand mutations with ~8% client aborts, zero errors with no aborts.

BlankParticle:
fixes #370 
premature client disconnects now don't reject with `TypeError: Body is unusable` if full body was received
for partial body, the rejection message is also made clearer

The test case and initial patch was written by Claude Fable 5 (medium)
I reviewed the code and test cases, manually tested, rewrote some of the stuff and added the normalised abort message

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -261,6 +261,40 @@ const readBodyWithFastPath = <T>(
   })
 }
 
+const normalizeAbortError = (
+  request: Record<string | symbol, any>,
+  incoming: IncomingMessage | Http2ServerRequest
+): Error => {
+  if (incoming.errored) {
+    return incoming.errored
+  }
+  const reason = request[abortReasonKey]
+  if (typeof reason !== 'undefined') {
+    return reason instanceof Error ? reason : new Error(String(reason))
+  }
+  return new Error('Client connection prematurely closed.')
+}
+
+// A client abort destroys the IncomingMessage, but does not clear its internal
+// buffer. For HTTP/1, `complete === true` guarantees the parser received the
+// entire message, so the unread body can still be drained synchronously.
+// (Not valid for HTTP/2, where `complete` is also true for aborted requests. see https://nodejs.org/api/http2.html#requestcomplete)
+const canBeRecoveredAfterAbort = (
+  incoming: IncomingMessage | Http2ServerRequest
+): incoming is IncomingMessage => {
+  return incoming.complete && incoming.httpVersionMajor === 1
+}
+
+const readIncomingMessageInternalBuffer = (incoming: IncomingMessage) => {
+  const chunks: Buffer[] = []
+  let chunk: Buffer | string | null
+  while ((chunk = incoming.read()) !== null) {
+    chunks.push(Buffer.isBuffer(chunk) ? chunk : Buffer.from(chunk))
+  }
+  const buffer = chunks.length === 1 ? chunks[0] : Buffer.concat(chunks)
+  return buffer
+}
+
 const readRawBodyIfAvailable = (request: Record<string | symbol, any>): Buffer | undefined => {
   const incoming = request[incomingKey] as IncomingMessage | Http2ServerRequest
   if ('rawBody' in incoming && (incoming as any).rawBody instanceof Buffer) {
@@ -283,7 +317,16 @@ const readBodyDirect = (request: Record<string | symbol, any>): Promise<Buffer> 
 
   const incoming = request[incomingKey] as IncomingMessage | Http2ServerRequest
   if (Readable.isDisturbed(incoming)) {
-    return rejectBodyUnusable()
+    if (incoming.readableDidRead) {
+      return rejectBodyUnusable()
+    }
+    // Aborted (e.g. client disconnect) before anything read the stream.
+    if (canBeRecoveredAfterAbort(incoming)) {
+      const buffer = readIncomingMessageInternalBuffer(incoming)
+      request[bodyBufferKey] = buffer
+      return Promise.resolve(buffer)
+    }
+    return Promise.reject(normalizeAbortError(request, incoming))
   }
 
   const promise = new Promise<Buffer>((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


usualoma:
Hi @BlankParticle,
Thanks for creating the pull request. I'm reviewing it by comparing it to what I have on my end, so please give me a moment.

usualoma:
Hi @BlankParticle,

Thank you for working on this. I confirmed that this PR fixes #370. Using the HTTP/1 parser's `complete` flag as a prerequisite for recovery is sound, and the change is nicely focused.

That said, a few behaviors make me hesitant to merge it as-is:

- If the raw stream is read after the disconnect (e.g. `incoming.read()` via `c.env.incoming`) before a body method is called, the recovery path resolves with an empty body instead of rejecting. On `main` this case rejects with `Body is unusable`, so this is a behavioral regression: a raw `read()` on a destroyed stream doesn't set `readableDidRead`, so the guard doesn't catch it and the now-empty internal buffer is accepted as the full body.
- The recovery path does not check `incoming.errored` before accepting the buffered body, so an application-provided stream error (e.g. a body-size guard calling `req.destroy(err)`) could be swallowed, causing the read to resolve instead of rejecting.
- When `setEncoding()` has been used, string chunks are converted back to bytes as UTF-8 regardless of the stream's actual encoding.
- The same disconnect scenario still fails through paths that bypass `readBodyDirect`, such as `formData()`, `c.req.raw.body`, and body reads that create or use the native Request cache after the disconnect.

I'd like to proceed with the fix from my existing branch instead, which also covers these cases and body-reading paths. I incorporated the parts of this PR that I found valuable in 138386d and 0007a09, and credited you as a co-author on both commits. The scenarios covered by this PR are covered on that branch as well.

I hope this approach works for you, and I'd be glad to have your review there. https://github.com/honojs/node-server/pull/375

BlankParticle:
Hi @usualoma,
thanks for checking on my PR, I made this without asking if you were working on it, I think it's fine if you can adapt parts of my PR to your existing branch. Once your PR is merged I will close this one.
I will check out and add review on your PR in a few minutes.

BlankParticle:
closed in favour of #375

PR #375, opened 2026-07-14
fix: recover complete request bodies after client disconnect

### LINKED ISSUE #370: v2: body read after client disconnect rejects with `TypeError: Body is unusable` even when the full body was received (regression from v1)

usualoma:
fixes #370

A client disconnect marks the `IncomingMessage` as disturbed even when Node's HTTP/1 parser has already received the complete request body. As a result, a later body read can incorrectly fail with `Body is unusable`.

This follows the same core approach as #373:

- Use `incoming.readableDidRead` instead of `Readable.isDisturbed(incoming)` to determine whether the body was actually consumed.
- If the HTTP/1 request was completed before the disconnect, recover the untouched body retained in the incoming stream's internal buffer.

The main difference is that the recovery is shared across the request body-reading paths rather than being limited to `readBodyDirect`. This also covers `formData()`, `c.req.raw.body`, and paths that create or use the native `Request` cache.

Although the diff is relatively large, the core behavior is the same. Most of the additional code and tests ensure that recovery is applied consistently while rejecting partially consumed bodies, preserving application-provided stream errors, and avoiding byte corruption when `setEncoding()` has been used.

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -407,6 +407,23 @@ const readRawBodyIfAvailable = (request: Record<string | symbol, any>): Buffer |
   return undefined
 }
 
+// The error a body read fails with after an abort: the stream's own error if
+// it has one, otherwise the abort reason recorded on the request, otherwise a
+// generic disconnect error.
+const normalizeAbortError = (
+  request: Record<string | symbol, any>,
+  incoming: IncomingMessage | Http2ServerRequest
+): Error => {
+  if (incoming.errored) {
+    return incoming.errored
+  }
+  const reason = request[abortReasonKey]
+  if (reason !== undefined) {
+    return reason instanceof Error ? reason : new Error(String(reason))
+  }
+  return new Error('Client connection prematurely closed.')
+}
+
 // Read body directly from the IncomingMessage stream, bypassing Request object creation.
 // Precondition: the caller (listener.ts) must ensure that the IncomingMessage stream is
 // properly cleaned up (e.g. via incoming.resume()) when the response ends or the connection
@@ -468,7 +485,7 @@ const readBodyDirect = (request: Record<string | symbol, any>): Promise<Buffer> 
         if (recovered instanceof Error) {
           reject(recovered)
         } else if (recovered === undefined) {
-          reject(error ?? new Error('Client connection prematurely closed.'))
+          reject(error ?? normalizeAbortError(request, incoming))
         } else {
           request[bodyBufferKey] = recovered
           resolve(recovered)
@@ -504,16 +521,7 @@ const readBodyDirect = (request: Record<string | symbol, any>): Promise<Buffer> 
         return
       }
       finish(() => {
-        if (incoming.errored) {
-          reject(incoming.errored)
-          return
-        }
-        const reason = request[abortReasonKey]
-        if (reason !== undefined) {
-          reject(reason instanceof Error ? reason : new Error(String(reason)))
-          return
-        }
-        reject(new Error('Client connection prematurely closed.'))
+        reject(normalizeAbortError(request, incoming))
       })
     }
     const cleanup = () => {
--- a/src/request.ts
+++ b/src/request.ts
@@ -68,11 +68,41 @@ const isByteExactEncoding = (encoding: BufferEncoding | null): boolean =>
   encoding === null || byteExactEncodings.has(encoding)
 
 const bodyBufferedBeforeDisconnectKey = Symbol('bodyBufferedBeforeDisconnect')
+const bodyBufferedLengthBeforeDisconnectKey = Symbol('bodyBufferedLengthBeforeDisconnect')
+
+type IncomingWithBodyRecovery = IncomingMessage & {
+  [bodyBufferedBeforeDisconnectKey]?: Buffer | Error
+  [bodyBufferedLengthBeforeDisconnectKey]?: number
+}
 
 // When setEncoding() was called on the stream, chunks arrive as strings in
 // that encoding — decode with it so the original bytes are reconstructed.
 const toBufferChunk = (chunk: Buffer | string, encoding: BufferEncoding | null): Buffer =>
   Buffer.isBuffer(chunk) ? chunk : Buffer.from(chunk, encoding ?? 'utf8')
+
+const isRecoverableDisconnectedIncoming = (
+  incoming: IncomingMessage | Http2ServerRequest
+): incoming is IncomingMessage =>
+  !(incoming instanceof Http2ServerRequest) &&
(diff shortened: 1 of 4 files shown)


src/request.ts
```diff
@@ -57,6 +57,128 @@ const newHeadersFromIncoming = (incoming: IncomingMessage | Http2ServerRequest)
...
+
+// When setEncoding() was called on the stream, chunks arrive as strings in
+// that encoding — decode with it so the original bytes are reconstructed.
+const toBufferChunk = (chunk: Buffer | string, encoding: BufferEncoding | null): Buffer =>
+  Buffer.isBuffer(chunk) ? chunk : Buffer.from(chunk, encoding ?? 'utf8')
+
+const isRecoverableDisconnectedIncoming = (
+  incoming: IncomingMessage | Http2ServerRequest
+): incoming is IncomingMessage =>
+  !(incoming instanceof Http2ServerRequest) &&
```
BlankParticle, line 86: not a fan of `instanceof` checks
because the input is either a http/1 or http/2 request, I think `incoming.httpVersionMajor === 1` is a much better check

src/request.ts
```diff
@@ -299,8 +463,39 @@ const readBodyDirect = (request: Record<string | symbol, any>): Promise<Buffer>
...
 
+    // readBodyDirect started while the stream was untouched, so it owns every
+    // chunk emitted from that point onward. If HTTP/1 parsing completed before
+    // the connection closed, combine those chunks with anything still buffered
+    // and treat the body as complete. Transport-level ECONNRESET is recoverable;
+    // application-provided stream errors are not.
+    const recoverCompleteBodyAfterDisconnect = (error?: unknown): boolean => {
+      const streamError = incoming.errored ?? error
+      if (
+        incoming instanceof Http2ServerRequest ||
```
BlankParticle: we should use the existing `isRecoverableDisconnectedIncoming` here and add the last condition with it

src/request.ts
usualoma: While I do understand your preference, since we've basically been using `incoming instanceof Http2ServerRequest` throughout this project, I think it would be more appropriate to use `incoming instanceof Http2ServerRequest` here as well.

BlankParticle: if it's a pattern already being used here, then I guess it's fine doing it here too
I just have a distrust of instanceof checks because libs including hono replaces the globals and these kind of checks fail in application code

src/request.ts
usualoma: Thank you. I've incorporated it. c86c3e1

BlankParticle (approved):

yusukebe (approved):
LGTM!

yusukebe:
@usualoma @BlankParticle Thank you both! I'll merge and release a new version.

PR #377, opened 2026-07-16
test: use a custom helper for path traversal tests

BlankParticle:
The test that were testing path traversal using `supertest` were not testing the proper behaviour. `supertest` uses `superagent` under the hood which automatically normalizes the path and there is no way to turn this behaviour off. This means hono was receiving normalized `/secret.txt` instead of `/static/../secret.txt`

`superagent` is full of some of these footguns, like in #368 `superagent` suddenly started doing brotli decompression in the major version without anyway to opt out, but it doesn't do zst compression. Now the test suite fixtures are testing against are 2 different types, one real brotli fixture and one fake zst fixture. 

https://e18e.dev is trying to get the ecosystem to migrate away from superagent https://github.com/e18e/ecosystem-issues/issues/257, so probably migrating away from `supertest` also makes sense

This PR introduces a small `requestServer` util that has similar interface to `fetch` built on node:http, this interface can replace like 90% of `supertest` usage in here, for the remaining we will need to add some specialized functions for things like testing HTTP/2 and TLS.

I am only replacing the affected path tests in this PR and introducing this new helper to the codebase as a cataleyist for the future migration

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -43,7 +43,7 @@ export class Request extends GlobalRequest {
   }
 }
 
-const newHeadersFromIncoming = (incoming: IncomingMessage | Http2ServerRequest) => {
+export const newHeadersFromIncoming = (incoming: IncomingMessage | Http2ServerRequest) => {
   const headerRecord: [string, string][] = []
   const rawHeaders = incoming.rawHeaders
   for (let i = 0, len = rawHeaders.length; i < len; i += 2) {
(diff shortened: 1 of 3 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@BlankParticle 

> `supertest` uses `superagent` under the hood which automatically normalizes the path and there is no way to turn this behaviour off. 

I see! Thanks.

PR #379, opened 2026-07-20
test: replace supertest

BlankParticle:
Follow up to #377

- Replaces `supertest` for custom helper functions
- Add custom helper functions to allow HTTP/2, chunk and header hooks and automatic lifecycle management of the passed in server
- Fix typescript issues by replacing depricated `Node` target for `Bundler` and module `EsNext` as the package is already being bundled by tsdown
- Reverted brotli fixtures back to fake compression before `supertest` bump to make it same as zst fixture

This was done in with co-ordination with GPT 5.6 Sol (medium), I overlooked the api surface and other things, the tedious job of porting over the tests to new interface was done by gpt. I along with gpt 5.6 and claude fable 5 reviewed the changes to make sure the test behaviours were kept same through out.

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,6 @@
   "devDependencies": {
     "@hono/eslint-config": "^1.0.1",
     "@types/node": "^20.10.0",
-    "@types/supertest": "^2.0.12",
     "@types/ws": "^8.18.1",
     "@whatwg-node/fetch": "^0.9.14",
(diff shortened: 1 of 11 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@BlankParticle 

Indeed, `supertest` was not really needed, and you did well. Thanks!

PR #382, opened 2026-07-24
fix(response): copy headers when init is a foreign Response

yusukebe:
I realized that the problem with rebuilding `Response` with a foreign Response as one returned from `fetch` will throw a ` TypeError: immutable`, is on the `@hono/node-server` side. The problem is pointed out https://github.com/honojs/hono/pull/5137#issuecomment-5017991712

This PR fixes it.

### DIFF
--- a/src/response.ts
+++ b/src/response.ts
@@ -46,21 +46,16 @@ export class Response {
   constructor(body?: BodyInit | null, init?: ResponseInit) {
     let headers: HeadersInit | undefined
     this.#body = body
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @usualoma 

Can you review this?

usualoma:
@yusukebe
I see, I understand! Thanks for the follow. LGTM!

yusukebe:
@usualoma Thanks!

PR #383, opened 2026-07-26
fix(listener): avoid uncaught error when force-closing a non-standard socket

### LINKED ISSUE #374: getRequestListener: uncaught `socket.destroySoon is not a function` in forceClose when the socket is mocked (e.g. fastify.inject via MCP SDK)
### What happened

`getRequestListener`'s request-drain logic force-closes the socket after `DRAIN_TIMEOUT_MS` when the incoming body was not fully read. `forceClose` assumes the socket is a real `net.Socket`:

```js
const forceClose = () => {
  cleanup();
  const socket = incoming.socket;
  if (socket && !socket.destroyed) {
    socket.destroySoon();
  }
};
```

When `incoming` is not a real `http.IncomingMessage` (for example a mocked request), `socket.destroyed` is `undefined` (so the guard passes) and `socket.destroySoon` does not exist, producing an **uncaught exception on a timer** that callers cannot catch:

```
TypeError: socket.destroySoon is not a function
 ❯ Timeout.forceClose node_modules/@hono/node-server/dist/index.mjs:390:14
 ❯ listOnTimeout node:internal/timers:585:17
 ❯ processTimers node:internal/timers:521:7
```

### How we hit this

The MCP TypeScript SDK (`@modelcontextprotocol/sdk`, `StreamableHTTPServerTransport`) uses `getRequestListener` to bridge Node req/res to fetch Request/Response. We mount that transport inside a Fastify route and integration-test it with `fastify.inject()` (light-my-request). light-my-request's `MockSocket` is a plain `EventEmitter` with neither `destroyed` nor `destroySoon`, so every injected MCP request schedules a 500 ms timer that later crashes the (vitest) worker with the unhandled `TypeError` above. Because it fires on a timer, it is timing-dependent and shows up as flaky "Unhandled Errors" in otherwise green test runs.

### Environment

- `@hono/node-server` 1.19.14 (via `@modelcontextprotocol/sdk` 1.29.0)
- Node v22.23.0
- light-my-request 6.6.0 (fastify inject)

### Suggested fix

Make `forceClose` defensive about non-standard sockets:

```js
const forceClose = () => {
  cleanup();
  const socket = incoming.socket;
  if (socket && !socket.destroyed) {
    if (typeof socket.destroySoon === 'function') {
      socket.destroySoon();
    } else {
      socket.destroy?.();
    }
  }
};
```

That keeps the drain behaviour for real sockets and turns the mocked-socket case into a no-op instead of an uncatchable crash. Happy to send a PR if that direction works for you.

mohamedramadan14:
Closes #374.

`forceClose` guards with `!socket.destroyed`, which is `undefined` on any socket that is not a real `net.Socket` — so the guard passes and `socket.destroySoon()` throws. It runs on a timer, so the `TypeError` becomes an `uncaughtException` the caller cannot catch.

This feature-detects the teardown method instead:

```ts
if (typeof socket.destroySoon === 'function') {
  socket.destroySoon()
} else if (typeof socket.destroy === 'function') {
  socket.destroy()
}
```

`net.Socket` and `tls.TLSSocket` both define `destroySoon`, so real connections keep the existing graceful-close behaviour unchanged; the check only ever diverges for sockets that could not have serviced the call. `cleanup()` still runs first on every path, so the timer and listeners are released regardless of which branch is taken.

This covers both entry points into `forceClose` — the `DRAIN_TIMEOUT_MS` timer and the `MAX_DRAIN_BYTES` overrun.

Tests drive `getRequestListener` with a mocked request whose socket is a bare `EventEmitter`, matching light-my-request's `MockSocket`: one asserts the drain timeout no longer throws, the other that a socket exposing only `destroy` is still torn down.

### DIFF
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -65,7 +65,11 @@ const drainIncoming = (incoming: IncomingMessage | Http2ServerRequest): void => 
     cleanup()
     const socket = incoming.socket
     if (socket && !socket.destroyed) {
-      socket.destroySoon()
+      if (typeof socket.destroySoon === 'function') {
+        socket.destroySoon()
+      } else if (typeof socket.destroy === 'function') {
+        socket.destroy()
+      }
     }
   }
 
--- a/test/listener.test.ts
+++ b/test/listener.test.ts
@@ -1,4 +1,7 @@
+import { EventEmitter } from 'node:events'
 import { createServer } from 'node:http'
+import type { IncomingMessage, ServerResponse } from 'node:http'
+import { Readable } from 'node:stream'
 import { getRequestListener } from '../src/listener'
 import { GlobalRequest, Request as LightweightRequest, RequestError } from '../src/request'
 import { GlobalResponse, Response as LightweightResponse } from '../src/response'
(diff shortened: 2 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@mohamedramadan14 Thanks!

PR #391, opened 2026-08-14
chore: add better benchmarks

BlankParticle:
Added a better benchmarks with extra cases inspired by srvx's nodejs compatibly suite, its also using `oha` instead of `bombardier` as `srvx` does. It uses more runs and measures in random order to account for random noise/activity, a full run takes about 10 minutes, `--update` automatically update the benchmark readme, there is a command provided to quickly smoke test the benchmark. We compare `@hono/node-server` (npm and local) and `srvx` against a baseline `node:http` version to measure overhead of the fetch servers in various scenarios.

---
> Disclosure: These changes were created by GPT 5.6 Sol, by looking at hono node-server and srvx's repo under my supervision

### DIFF
--- a/benchmarks/fetch/README.md
+++ b/benchmarks/fetch/README.md
@@ -1,13 +1,13 @@
-# Benchmark
+# Fetch server benchmark
 
-Benchmark comparing the published npm version and local development version of @hono/node-server with srvx.
+This suite measures the raw HTTP-to-Fetch adapter cost of the published and development versions of `@hono/node-server`. A native `node:http` implementation is the performance floor, while srvx with `FastResponse` is another Fetch-compatible adapter for comparison.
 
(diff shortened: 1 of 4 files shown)


yusukebe:
Hey @BlankParticle 

This is great! Can we also have a single ranked table like [srvx has](https://github.com/h3js/srvx/tree/main/test/bench-node)? Something like "one number per server, fastest first". Like the below:

CleanShot 2026-08-16 at 22 07 32@2x

BlankParticle:
hello @yusukebe,
looks like srvx uses json round trip as the benchmark, i have added the section and it compares against node http as the baseline

yusukebe (approved):
LGTM!

yusukebe:
> looks like srvx uses json round trip as the benchmark, i have added the section and it compares against node http as the baseline

I see! Looks good. Thanks @BlankParticle

PR #393, opened 2026-08-17
docs(readme): update benchmark results

Jr-14:
Syncing the README and benchmark results from https://github.com/honojs/node-server/pull/391 into the root README :)

I've removed the express benchmark - not sure if it's worth keeping given express is now in v5. Probably would be a good idea to re-create the benchmarks?

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,35 +5,30 @@ Initially, Hono wasn't designed for Node.js, but with this adapter, you can now 
 
 ## Benchmarks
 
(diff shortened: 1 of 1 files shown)


yusukebe:
@Jr-14 

This change is not needed. We list a benchmark comparing with only Express.

yusukebe:
I updated the README https://github.com/honojs/node-server/pull/394

PR #397, opened 2026-08-25
fix(request): release cloned request bodies via buffer-once replay

### LINKED ISSUE #347: V2 Memory Leak ( solved: c.req.raw.clone().json() )
Hey there!

Noticed a memory leak in one of our services due to upgrading to V2 of hono node server.

After some lengthy profiling/general troubleshooting, we determined the issue was stemming from our usage of `c.req.raw.clone().json()`; when removed, the memory leak went away.

Image *cert environment: before upgrade, during upgrade, after root cause fix (note: the spikes are nightly stress tests)*

Image *production: steady linear increase (with dropoffs every deploy) until fix put in place*

We've fixed it from our side, but wanted to create this issue for visibility.

```typescript
// truncated code of our before state for brevity (see: logger middleware at bottom)

export class GraphQLRoute extends Hono {
    constructor(db: DBClient) {
        super();

        this.on(['POST', 'GET'], GRAPHQL_ENDPOINT, async (c) => {
            const yoga = createYoga({
                schema: buildSchema(db),
                graphqlEndpoint: GRAPHQL_ENDPOINT,
            });

            return yoga.fetch(c.req.raw, c);
        });
    }
}

export class Server extends Hono {
    constructor(db: DBClient) {
        super();

            this.use(loggerMiddleware({logger}))
            .route('', new GraphQLRoute(db))
    }
}

export const loggerMiddleware = (opts: {logger: typeof logger}) => {
    return createMiddleware(async (c, next) => {
        const loggerContext: Partial<Record<LoggableField, LoggableValue>> = {
            payload: await c.req.raw.clone().json(), // we have to clone otherwise the stream will already have been read downstream and error
        };

        await opts.logger.runInContext(loggerContext, async () => {/* ... */});
    });
};

```

After narrowing down the code via datadog profiling, copilot had this explanation if it's worth anything:

> In @hono/node-server v2, the underlying Request body is a ReadableStream backed by the Node.js socket. Request.clone() calls ReadableStream.tee(), which buffers all chunks internally until both branches are consumed. The logger middleware consumes the clone branch immediately, but the original branch isn't consumed until Yoga/REST handlers run later (after next()). In v2, this tee buffer holds the full body in memory far longer than v1's implementation did — the +86 MiB the flame graph attributes to this file

SynthLuvr:
Fixes #347

## Problem

`clone()` tee()s the socket-backed body stream. The unread branch of a tee queues every chunk until it is read, so when middleware does `c.req.raw.clone().json()` and the original body is never read, the entire raw body stays resident for as long as the request graph is reachable — logger/tracing contexts, error objects, and profilers routinely retain the request past the response, which shows up as the linear memory growth reported in the issue.

## Approach

- `clone()` now reads the body exactly once into a Buffer (via the existing `readBodyDirect` fast path) and hands the original request and every clone a lightweight replay stream backed by that shared buffer. No tee is created, so nothing queues an unread copy.
- When the response completes, the listener drops the shared buffer and cancels replay streams nobody is reading. Locked streams that are actively being read are left alone and finish naturally. A retained request therefore no longer pins the raw body.
- The release hook on the synchronous cacheable fast path listens on `'close'` rather than `'finish'` so the buffer is also released when the client disconnects before the response is flushed — and it is only attached when a shared body exists, so responses without cloned bodies keep the path free of close listeners.

Reading an unread clone *after* the response has completed rejects with `TypeError: Body is unusable`, matching what a native `Request` does when its body stream has been canceled.

## Tests

- `test/request.test.ts`: clones replay the body while leaving the original readable; unread clones are canceled on release; clones being read keep delivering
- `test/listener.test.ts`: the original body stays readable after a clone has consumed it (integration); the synchronous cacheable fast path releases the shared body on close
- `test/server.test.ts`: GC-based memory-leak regression next to the existing ones — the shared buffer is collected after the response completes even while the request is still retained

Verified locally: `format`, `lint`, `typecheck`, `build` + publint, and the full vitest suite.

### DIFF
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -5,6 +5,8 @@ import type { Writable } from 'node:stream'
 import type { IncomingMessageWithWrapBodyStream } from './request'
 import {
   abortRequest,
+  bodySharedBufferKey,
+  releaseSharedBody,
   newRequest,
   recordBodyBufferedBeforeDisconnect,
   Request as LightweightRequest,
@@ -99,6 +101,8 @@ const makeCloseHandler =
     needsBodyCleanup: boolean
   ): (() => void) =>
   () => {
+    // The response is over: release clone-shared body buffers.
+    req[releaseSharedBody]()
     if (incoming.errored) {
       recordBodyBufferedBeforeDisconnect(incoming)
       req[abortRequest](incoming.errored.toString())
@@ -431,6 +435,16 @@ export const getRequestListener = (
         // Synchronous cacheable response — no close listener needed.
         // No I/O events can fire between fetchCallback returning and responseViaCache
         // completing, so abort detection is not needed here.
+        // Release the clone-shared body buffer once the response completes, but
+        // only when there is one: the fast path must stay free of close
+        // listeners otherwise. Hook 'close' rather than 'finish' so the buffer
+        // is also released when the client disconnects before the response is
+        // flushed.
+        if (req[bodySharedBufferKey]) {
+          outgoing.once('close', () => {
+            req[releaseSharedBody]()
+          })
+        }
         if (needsBodyCleanup && !incoming.readableEnded) {
           // Handler returned without consuming the body; drain after the
           // response is flushed so the socket is freed gracefully (avoids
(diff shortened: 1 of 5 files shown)


yusukebe:
We close https://github.com/honojs/node-server/pull/397. This PR is not necessary.

SynthLuvr:
> We close https://github.com/honojs/node-server/pull/397. This PR is not necessary.

@yusukebe what do you mean?

yusukebe:
@SynthLuvr 

Sorry this issue: https://github.com/honojs/node-server/issues/347

PR #399, opened 2026-09-07
perf(response): defer streams when copying response headers

colinhacks:
Preserve buffered string/null bodies when Hono copies a finalized response to update its headers. The optional constructor hook copies headers without constructing a ReadableStream, and the cached writer records consumption so saved response aliases cannot replay the body.

Body observation materializes native responses sharing the same stream. Unsupported responses use the existing path, including streams, mutable byte arrays, subclasses, proxies and already-observed bodies. Shared groups are capped at 64 wrappers.

The companion Hono change, https://github.com/honojs/hono/pull/5350, calls the hook before reading the body. Without it, the adapter retains the existing behavior. No application changes or runtime-specific condition are required.

### Performance

Runnable reproduction, custom fixtures and raw results: [hono-response-copy-benchmark](https://github.com/colinhacks/hono-response-copy-benchmark/tree/c171898). The repository builds the exact source revisions below and includes a second clean-VM verification run. These are custom fixtures, not Hono's official benchmark suite.

**With both PRs applied, mixed-route throughput improved 65%, eight-header throughput 62%, and 64 KiB response throughput 31%.** Application code was identical across controls. These gains require both the Hono hook and the adapter implementation; neither PR alone enables the fast path.

Median requests/second:

| Workload | Unmodified packages | Both PRs | Both PRs, hook disabled | Change vs unmodified |
| --- | ---: | ---: | ---: | ---: |
| Mixed routes and middleware | 16,254 | 26,857 | 15,891 | +65% |
| Eight post-handler headers | 11,086 | 18,000 | 10,841 | +62% |
| 64 KiB string with middleware | 8,086 | 10,625 | 7,511 | +31% |
| Plain text, no middleware | 39,897 | 39,223 | 39,516 | −1.7% |

Node 26.8.1 on a dedicated Linux GCE n2-standard-8 VM. HTTP/1.1 loopback, 32 connections, two wrk threads, server and client pinned to separate physical cores. Three interleaved repetitions, each with a two-second warmup and five-second measurement. Mixed traffic was 80% middleware-backed parameter routes across 16 routes and 4,096 IDs, 10% text and 10% JSON. At two connections, mixed throughput improved from 14,973 to 25,679 requests/second (+71%).

The comparison used source-built Hono `eebdf7b` and adapter `64dc09e` as controls, versus Hono `b0855ef` and adapter `066bf86`. Subsequent commits add tests only. These are short, closed-loop fixture measurements, not production latency or application-wide performance claims.

### Tests

- Added header-copy, alias-consumption, fallback and real HTTP GET/HEAD regressions.
- Full suite: 448 passed on Node 20 and 22; 449 passed on Node 22 after adding the Content-Type regression. Typecheck, lint, format and ESM/CJS build passed.
- Cross-package differential checks passed on Node 20, 22, 24 and 26; CJS integration passed.
- Full suites on Node 24/26 retained one/three request partial-read or disconnect failures, respectively, reproduced on the unchanged base with the same dependencies.

### DIFF
--- a/src/listener.ts
+++ b/src/listener.ts
@@ -11,7 +11,12 @@ import {
   wrapBodyStream,
   toRequestError,
 } from './request'
-import { defaultContentType, cacheKey, Response as LightweightResponse } from './response'
+import {
+  defaultContentType,
+  cacheKey,
+  consumeSharedBody,
+  Response as LightweightResponse,
+} from './response'
 import type { InternalCache } from './response'
 import type { CustomErrorHandler, FetchCallback, HttpBindings } from './types'
 import {
@@ -169,6 +174,9 @@ const responseViaCache = async (
   res: Response,
   outgoing: ServerResponse | Http2ServerResponse
 ): Promise<undefined | void> => {
+  if (!consumeSharedBody(res)) {
+    return responseViaResponseObject(res, outgoing)
+  }
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   let [status, body, header] = (res as any)[cacheKey] as InternalCache
 
--- a/src/response.ts
+++ b/src/response.ts
@@ -2,12 +2,14 @@
 // Define lightweight pseudo Response class and replace global.Response with it.
 
 import type { OutgoingHttpHeaders } from 'node:http'
+import { types } from 'node:util'
 
 export const defaultContentType = 'text/plain; charset=UTF-8'
 
 const responseCache = Symbol('responseCache')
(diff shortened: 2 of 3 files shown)


colinhacks:
The Node 24 failure is in the existing partial-body test (`request.test.ts:318`), reproduced on the unchanged base `64dc09e` as well. Neither the request implementation nor that test changes here. All 14 response-copy tests passed; the Node 20/22 jobs were cancelled by matrix fail-fast. The Windows suite and package build passed.

